PR #6, opened 2021-07-06
boilerplate comment is not mandatory

smaeda-ks:
https://developer.fastly.com/reference/vcl/subroutines/#concatenation

> Built-in subroutines with the same name are automatically concatenated in the order the compiler encounters them.

There are valid use cases to not include Fastly's code macros if it's intended to be concatenated. Therefore, Severity should be WARNING.


ysugimoto (approved):
Thanks, it would be nice 👍

PR #9, opened 2021-07-06
skip duplication check against Fastly's reserved subroutines

smaeda-ks:
Similar to: https://github.com/ysugimoto/falco/pull/6

### DIFF
--- a/context/context.go
+++ b/context/context.go
@@ -2,7 +2,6 @@ package context
 
 import (
 	"fmt"
-	"regexp"
 	"strings"
 
 	"github.com/ysugimoto/falco/types"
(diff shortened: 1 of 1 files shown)


context/context.go
```diff
@@ -45,6 +46,12 @@ func ScopeString(s int) string {
 	}
 }
 
+var fastlySubRegex = regexp.MustCompile("^vcl_(recv|hash|hit|miss|pass|fetch|error|deliver|log)$")
+
+func IsFastlySubroutine(name string) bool {
+	return fastlySubRegex.MatchString(name)
+}
+
```
ysugimoto, line 54: The regular expression in Go sometimes slow (frequently talked about this 🙃 ) and Fastly's reserved subroutines do not have too many patterns.

How about using a defined map and search from it?

```suggestion
var fastlyReservedSubroutines = map[string]struct{}{
	"vcl_recv": {},
	"vcl_hash": {},
	"vcl_hit": {},
	"vcl_miss": {},
	"vcl_pass": {},
	"vcl_fetch": {},
	"vcl_error": {},
	"vcl_deliver": {},
	"vcl_log": {},
} 

func IsFastlySubroutine(name string) bool {
	_, ok :=  fastlyReservedSubroutines[name]
	return ok
}
```

But this is a tiny thing and your change looks neat ✨   it's up to you 👍

ysugimoto (approved):
I understand that your change comes from here: https://developer.fastly.com/reference/vcl/subroutines/#concatenation

Thank you for that!

context/context.go
smaeda-ks: Thanks! Pushed another commit which replaces regexp with `map[string]bool`.

PR #22, opened 2021-07-12
fix add statement linter

ysugimoto:
This PR has a couple of changes:

1. Change severity from `WARNING` to `ERROR` when type is different between left and right on `=` operator in `set` statement
2. Raise `ERROR` when `add` statement is used not for HTTP headers (refers [add statement](https://developer.fastly.com/reference/vcl/statements/add/)
3. Check validity of regex string inside if expression


smaeda-ks (approved):
LGTM :+1:

PR #23, opened 2021-07-13
use context.Set() for set statement linter

smaeda-ks:
`context.Get()` shouldn't be used in `lintSetStatement` as that will skip nesesarry validations in `context.Set()`.

```
sub vcl_fetch {
  set beresp.proto = "http";
  set beresp.saintmode = 10s;
}
```

### expected

For instance, the above example should throw an error for `set beresp.proto` as it's a read-only variable. And there should be no errors for `set beresp.saintmode = 10s;`.

```
% ./falco foo.vcl
🔥 [ERROR] variable "beresp.proto" is read-only
document: https://developer.fastly.com/reference/vcl/variables/backend-response/beresp-proto/
in /path/to/foo.vcl at line 2, position 7
 1|sub vcl_fetch {
 2|  set beresp.proto = "http";
         ^^^^^^^^^^^^
 3|  set beresp.saintmode = 10s;

🔥 [ERROR] beresp.proto wants type NULL but assign STRING
in /path/to/foo.vcl at line 2, position 22
 1|sub vcl_fetch {
 2|  set beresp.proto = "http";
                        ^^^^^^
 3|  set beresp.saintmode = 10s;

🔥 2 errors, ❗️ 1 warnings, 🔈 0 infos.
```

### actual

```
 $ ./falco foo.vcl
🔥 [ERROR] variable "beresp.saintmode" could not read
document: https://developer.fastly.com/reference/vcl/variables/backend-response/beresp-saintmode/
in /path/to/foo.vcl at line 2, position 7
 6|  set beresp.proto = "http";
 7|  set beresp.saintmode = 10s;
         ^^^^^^^^^^^^^^^^
 8|}

🔥 [ERROR] beresp.saintmode wants type NULL but assign RTIME
in /path/to/foo.vcl at line 2, position 26
 6|  set beresp.proto = "http";
 7|  set beresp.saintmode = 10s;
                            ^^^
 8|}

🔥 2 errors, ❗️ 1 warnings, 🔈 0 infos.
```

### DIFF
--- a/context/context.go
+++ b/context/context.go
@@ -288,17 +288,17 @@ func (c *Context) Get(name string) (types.Type, error) {
 	return obj.Value.Get, nil
 }
 
-func (c *Context) Set(name string, value types.VCLType) error {
+func (c *Context) Set(name string) (types.Type, error) {
 	first, remains := splitName(name)
 
 	// regex group variable like "re.group.N" is known read-only,
 	if first == "re" {
-		return fmt.Errorf(`variable "%s" is read-only`, name)
+		return types.NullType, fmt.Errorf(`variable "%s" is read-only`, name)
 	}
 
 	obj, ok := c.Variables[first]
 	if !ok {
-		return fmt.Errorf(`undefined variable "%s"`, name)
+		return types.NullType, fmt.Errorf(`undefined variable "%s"`, name)
 	}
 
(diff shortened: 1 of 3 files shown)


ysugimoto (approved):
LGTM 👍 Thanks!

PR #24, opened 2021-07-13
revert regex related linting

ysugimoto:
My apologies that Varnish regular expression is not compatible for Golang's regexp package.

It causes:

```
if (req.url ~ "^/([^\?]*)?(\?.*)?$") {
...
}
```

Then raises **[ERROR] regex string is invalid, error parsing regexp: invalid or unsupported Perl syntax: (?.**
I assume that VCL regular expression is using Perl syntax but Golang does not have it... so I revert these changes 😢

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.16
 
 require (
 	github.com/fatih/color v1.12.0
-	github.com/goccy/go-yaml v1.8.9 // indirect
+	github.com/goccy/go-yaml v1.8.9
 	github.com/google/go-cmp v0.5.6
 	github.com/k0kubun/colorstring v0.0.0-20150214042306-9440f1994b88 // indirect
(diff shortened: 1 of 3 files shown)


smaeda-ks (approved):

ysugimoto:
Temporary revert this change. I guess it may fix by escaping back-slash. 
`"^/([^\?]*)?(\?.*)?$"` does not seem Perl syntax, just matches `?` character 🤔

smaeda-ks:
I think you're right. Can `regexp.QuoteMeta` help here?
https://pkg.go.dev/regexp#QuoteMeta

ysugimoto:
Thanks, I think that function also escapes other meta characters like `[`.
Simply we can replace escape back-slack like `strings.ReplaceAll(regexString, "\\", "\\\\")`.

smaeda-ks:
Ah, I misunderstood the point. Sounds good.

PR #27, opened 2021-08-04
Fix definition of TCP congestion parameters

bungoume:
* [client.socket.congestion_algorithm](https://developer.fastly.com/reference/vcl/variables/client-connection/client-socket-congestion-algorithm/)
* [client.socket.cwnd](https://developer.fastly.com/reference/vcl/variables/client-connection/client-socket-cwnd/)

These settings are configurable.


ysugimoto (approved):
LGTM 👍  Thanks!

PR #32, opened 2021-11-03
Rename main package to cmd/falco

xordspar0:
The default binary name is the name of the package it comes from.
Previously, running `go install github.com/falco/cli` would install a
binary called `cli` to GOBIN. We want the main package to be called
falco.


ysugimoto (approved):
Nice work, thanks!

PR #43, opened 2022-03-01
Fixes #42 - Linting `client.geo.city.utf8`

### LINKED ISSUE #42: [BUG] Parser/Linter error `client.geo.city.utf8` is not supported
Hey folks,

Awesome project, I am testing it out for our vcl :)

**Describe the problem**

 I encountered the following issue:
 
 ```
 diff --git a/examples/default01.vcl b/examples/default01.vcl
index ad98204..d3e1be4 100644
--- a/examples/default01.vcl
+++ b/examples/default01.vcl
@@ -39,7 +39,8 @@ sub vcl_deliver {
 }
 
 sub vcl_fetch {
-
+  log client.geo.city;
+  log client.geo.city.utf8;
   #Fastly fetch
   return(deliver);
 }
 ```
 
 Apply the patch above and see that  [client.geo.city.utf8](https://developer.fastly.com/reference/vcl/variables/geolocation/client-geo-city-utf8/) is not recognized by the linter and that leads to errors.
 
 I am not sure how this happens since the variable is added to `predefined.yaml`. I think that these might be the only variables that have depth 4.
 
 For completeness it works on fastly fiddle [link](https://fiddle.fastlydemo.net/fiddle/b885e2cd)

davinci26:
As discussed in #42, setting the split depth to 3 causes variables
such as `client.geo.city.utf8` to not be recognized.

### DIFF
--- a/context/context.go
+++ b/context/context.go
@@ -518,7 +518,7 @@ func splitName(name string) (string, []string) {
 	var first string
 	var remains []string
 
-	sep := strings.SplitN(name, ".", 3)
+	sep := strings.SplitN(name, ".", 4)
 	if len(sep) == 1 {
 		first = sep[0]
 	} else {
--- a/context/context.go
+++ b/context/context.go
@@ -396,6 +396,13 @@ func (c *Context) Declare(name string, valueType types.Type, m *ast.Meta) error 
 
 	first, remains := splitName(name)
 
+	// declaration syntax for variables is:
+	// declare local var.variableName [type]
+	// which means that they must be prefixed with var.
+	if first != "var" {
+		return fmt.Errorf(`variable "%s" declaration error. Variable be prefixed with 'var.'`, name)
+	}
+
 	obj, ok := c.Variables[first]
 	if !ok {
 		// Newly assign object
--- a/context/context_test.go
+++ b/context/context_test.go
@@ -130,6 +130,9 @@ func TestContextDeclare(t *testing.T) {
 		if err := c.Declare("var.foo", types.StringType, nil); err == nil {
 			t.Errorf("expected error but got nil")
 		}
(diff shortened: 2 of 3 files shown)


ysugimoto (approved):
@davinci26 Thanks for your PR, It looks very nice!

And... I'd like to request additional change 😃 

----

 The `splitName` utility function is used for all variable splitting and also used in `Declare` which is a local variable declaration [here](https://github.com/davinci26/falco/blob/fixClientIp/context/context.go#L397) .

However, the local variable declaration syntax should be:

```
declare local var.variableName [type]
```

then split names should be validated in a variable declaration, `first` is `var`, and `remains` count is 1.

There is no validation yet, so if you could add this validation, the linter logic becomes more robust and it really helps us!

But, this is not related to your issue, so feel free to implement it. Your PR is ready to merge!

davinci26:
yeah I can add that test, no prob

davinci26:
I am looking into this code and I think this might be implemented in `statement_parser_test.go`
```
func TestDeclareStatement(t *testing.T) {
	input := `// Subroutine
sub vcl_recv {
	// Leading comment
	declare local var.foo STRING; // Trailing comment
}`
```

am I missing something here?

davinci26:
Should `remains count be 1.`? I see that fastly accepts local var.foo.bar here https://fiddle.fastlydemo.net/fiddle/b93197e6

ysugimoto:
Hmm, I didn't know that.
Fastly's documentation does not say that syntax, so it is an undocumented thing...
https://developer.fastly.com/reference/vcl/statements/declare/

Anyway, your additional change should be fine, let's ignore how `remains count` are.

ysugimoto:
@davinci26

This PR is ready to merge, I'll merge this and bump the patch version.
Thank you for your cooperation!

PR #64, opened 2022-05-04
feat(lint): Add warning for backend health check prober

davinci26:
If the backend prober has an initial value that is less than the threshold
then when the backend is loaded the initial health checks will not be sufficient
to mark it as healthy.

As such the backend will be considered unhealthy at startup.

I know that it doesnt follow the spec of VCL but it seems valuable to have this imo.

### DIFF
--- a/linter/helper.go
+++ b/linter/helper.go
@@ -3,6 +3,7 @@ package linter
 import (
 	"fmt"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/ysugimoto/falco/ast"
@@ -341,4 +342,39 @@ func hasFastlyBoilerPlateMacro(commentText, phrase string) bool {
 		}
 	}
 	return false
+}
+
+// According to fastly if a prober is configured with initial < threshold
+// then the prober will be marked as unhealthy at the beginning which is can
+// cause issues at startup.
+func isProbeMakingTheBackendStartAsUnhealthy(prober ast.BackendProbeObject) error {
+
+	var threshold int
+	var initial int
+	var err error
+	for _, v := range prober.Values {
+		if strings.EqualFold(v.Key.Value, "initial") {
+			// Optimistically cast this to int, if we can do it because the value is a literal
+			// we can apply the check
+			initial, err = strconv.Atoi(v.Value.String())
+			if err != nil {
+				return nil
+			}
+		}
+
+		if strings.EqualFold(v.Key.Value, "threshold") {
+			// Optimistically cast this to int, if we can do it because the value is a literal
+			// we can apply the check
+			threshold, err = strconv.Atoi(v.Value.String())
+			if err != nil {
+				return nil
+			}
+		}
+	}
+
+	if threshold > initial {
+		return fmt.Errorf("healthcheck initial value is lower than the threshold. Backend will start as unhealthy")
+	}
+
+	return nil
 }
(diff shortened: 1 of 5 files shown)


linter/rules.go
```diff
@@ -15,6 +15,7 @@ const (
 	BACKEND_SYNTAX                       = "backend/syntax"
 	BACKEND_DUPLICATED                   = "backend/duplicated"
 	BACKEND_NOTFOUND                     = "backend/notfound"
+	BACKEND_PROBER_CONFIGURATION         = "backend/prober-configuration"
```
shadialtarsha, line 18: I think adding a reference for this rule would be a good idea. 
If you check the example that Fastly provides in this [page](https://developer.fastly.com/reference/vcl/declarations/backend/) you can see that they provide an initial value less than the threshold which is confusing tbh for the user

linter/helper.go
```diff
@@ -342,3 +343,37 @@ func hasFastlyBoilerPlateMacro(commentText, phrase string) bool {
...
+		if strings.EqualFold(v.Key.Value, "threshold") {
+			// Optimistically cast this to int, if we can do it because the value is a literal
+			// we can apply the check
+			threshold, err = strconv.Atoi(v.Value.String())
+			if err != nil {
+				return nil
+			}
+		}
+	}
+
```
shadialtarsha, line 373: Do you think we need to care about the case where one of the options is not provided?

linter/rules.go
davinci26: What you get is:

```
sotiris.nanopoulos@C02CN3CMML7K falco % ./dist/falco-darwin-amd64 -v examples/default01.vcl
❗️ [WARNING] healthcheck initial value is lower than the threshold. Backend will start as unhealthy (backend/syntax)
in /Users/sotiris.nanopoulos/dev/falco/examples/default01.vcl at line 16, position 4
 15|  .max_tls_version = "1.2";
 16|  .probe = {
       ^^^^^
 17|    .request = "GET /status/200 HTTP/1.1" "Host: httpbin.org" "Connection: close";
See document in detail: https://developer.fastly.com/reference/vcl/declarations/backend/
```

davinci26: I think fastly example is just wrong in this case so it is hard to provide a reference example but I have observed this anecdotally

linter/helper.go
davinci26: I thought about this so I wrote the code in such a way that the following happens:

* if both dont appear then both have values 0 and 0 in which case you get no warning
* if initial exists but no threshold then it is also correct because initial value > 0
* if threshold exists but no initial then it is also correct because threshold > 0

davinci26: added a comment in e77ceccb2c4448bb01a0dfb3c1e11ed00d5c6791 to have this as part of the code as well

shadialtarsha (approved):
LGTM 🚀

davinci26:
Also fiddle doesnt allow for backends so this is why I havent shared fiddles with the behaviour

ysugimoto (approved):
I really want to lint such an implicit behavior.
This PR is incredibly nice, thanks!!

davinci26:
Awesome, glad you liked it, was worried a bit if it fitted the project! Yeah thats a goal for me to able to codify some knowledge that we have acquired into a code :)

PR #65, opened 2022-05-09
Add support to parse function calls

shadialtarsha:
**What is this PR?**
This PR adds support for calling functions like statements. 
An example for that would be calling [std.collect()](https://developer.fastly.com/reference/vcl/functions/miscellaneous/std-collect/)
```
# For a request with these Cookie headers:
# Cookie: name1=value1
# Cookie: name2=value2
std.collect(req.http.Cookie, ";");
# req.http.Cookie is now "name1=value1; name2=value2"
```

### DIFF
--- a/ast/functioncall_statement.go
+++ b/ast/functioncall_statement.go
@@ -1,0 +1,30 @@
+package ast
+
+import (
+	"bytes"
+)
+
+type FunctionCallStatement struct {
+	*Meta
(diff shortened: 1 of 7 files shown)


shadialtarsha:
I am not 100% happy about the implementation as there are some repeated code between functions as expression and functions as statements but due to the current design I implemented it like I did here.

PR #66, opened 2022-05-10
Support PCRE Regular Expressions

shadialtarsha:
**Why?**
Falco ([the go regexp pkg](https://pkg.go.dev/regexp) to be more accurate!) right now doesn't know how to parse [PCRE](https://www.pcre.org/) regular expressions that Varnish uses.
An example of that would be this file:
https://github.com/varnishcache/varnish-devicedetect/blob/master/devicedetect.vcl

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,17 +14,17 @@ darwin-deps:
 linux-deps:
 	apt-get install -y libpcre3-dev
 
-linux: linux-deps 
+linux:
 	GOOS=linux GOARCH=amd64 CGO_ENABLED=1 go build \
 			 -ldflags "-X main.version=$(BUILD_VERSION)" \
 			 -o dist/falco-linux-amd64 ./cmd/falco
 
-darwin_amd64: darwin-deps
+darwin_amd64:
 	GOOS=darwin GOARCH=amd64 CGO_ENABLED=1 go build \
 			 -ldflags "-X main.version=$(BUILD_VERSION)" \
 			 -o dist/falco-darwin-amd64 ./cmd/falco
 
-darwin_arm64: darwin-deps
+darwin_arm64:
 	GOOS=darwin GOARCH=arm64 go build \
 			 -ldflags "-X main.version=$(BUILD_VERSION)" \
 			 -o dist/falco-darwin-arm64 ./cmd/falco
--- a/Makefile
+++ b/Makefile
@@ -8,20 +8,23 @@ generate:
 test: generate
 	go list ./... | xargs go test
 
-linux:
+darwin-deps:
+	brew list pcre || brew install pcre
+
+linux-deps:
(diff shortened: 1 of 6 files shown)


shadialtarsha:
Tests need to be added 
I am on it

Makefile
```diff
@@ -27,7 +32,5 @@ lint:
 local: test lint
 	go build ./cmd/falco
 
-all: linux darwin
```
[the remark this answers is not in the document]
shadialtarsha: Right now `make all` wouldn't work anymore on mac because of the use of `apt-get` and I thought no point in keeping it, wdyt?

Makefile
```diff
@@ -9,14 +9,19 @@ test: generate
...
-	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build \
+	apt-get install -y libpcre3-dev
+	GOOS=linux GOARCH=amd64 CGO_ENABLED=1 go build \
 			 -ldflags "-X main.version=$(BUILD_VERSION)" \
 			 -o dist/falco-linux-amd64 ./cmd/falco
 
-darwin:
-	GOOS=darwin GOARCH=amd64 go build \
+darwin_amd64:
+	brew install pcre
```
davinci26, line 18: `brew install pcre` adds a significant overhead at every compilation because brews tries to update. is there a way we can avoid that?

ysugimoto:
@shadialtarsha How is it going?

About https://github.com/ysugimoto/falco/pull/66#discussion_r873408885 , Does the user need to install `pcre` when using `falco`?
If the user does not need to consider it, I think it's OK.

shadialtarsha:
That is the thing with `pcre`
It needs to be installed! 
I am gonna try to find a way to include the library in the codebase somehow without the need to install it! 
If you have also an idea, let me know!

davinci26:
I would make a `make deps` statement in the makefile and use that to conditionally import PCRE. If you want in this step you can test if `pcre` is already there and if not do nothing. 

For me as a mac user the cost is that each build step brew gets invoked which cases brew to do an update which takes a bunch of time

Makefile
```diff
@@ -8,15 +8,23 @@ generate:
 test: generate
 	go list ./... | xargs go test
 
-linux:
-	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build \
+darwin-deps:
+	brew list pcre || brew install pcre
```
shadialtarsha, line 12: This will install PCRE only if it is not installed

davinci26 (approved):
🐟

davinci26:
@ysugimoto are you ok with this approach?

ysugimoto:
I think it's gonna be OK.
Usually, a user uses a built binary for each platform so only the developer needs to consider it.

ysugimoto (approved):

PR #70, opened 2022-06-03
doc: Fix unexpected strikethrough is appeared on rendered HTML

arayaryoma:
In the current document, some `~` is treated as [Strikethrough](https://github.github.com/gfm/#strikethrough-extension-) expression of GFM.

I fixed to use [Code spans](https://github.github.com/gfm/#code-spans) to fix this.


ysugimoto (approved):
Thanks!

Issue #68, opened 2022-05-17
[Feature] Add Support for goto


shadialtarsha:
**Describe the feature you'd like**
Faslty supports `goto` syntax which Falco doesn't know how to lint.

Example:
```
sub test_sub {
  declare local var.x STRING;
  declare local var.y STRING;
  
  set var.x = "x";
  set var.y = "y";

  if (var.x == "x") {
      goto update_and_set;
  }

  // This code is not gonna be executed.
  set req.http.x-header1 = var.x;

  update_and_set:

  set req.http.x-header2 = var.y;
}
```

**Additional context**
Write some conditions, related links you've considered

ysugimoto:
@shadialtarsha I didn't know this feature.
Is it documented on Fastly docs?

shadialtarsha:
Unfortunately Fastly doesn't have documentation for it. 
Yes, it is a varnish feature.
I was also surprised when I saw it

PR #72, opened 2022-06-07
Add support for Goto declaration

shadialtarsha:
This PR adds support for goto declaration and use.
(Fixing https://github.com/ysugimoto/falco/issues/68)


linter/linter.go
```diff
@@ -1041,6 +1090,11 @@ func (l *Linter) lintIdent(exp *ast.Ident, ctx *context.Context) types.Type {
 			// mark table is used
 			t.IsUsed = true
 			return types.TableType
+		} else if t, ok := ctx.Gotos[exp.Value]; ok {
+			fmt.Println("herere")
```
ysugimoto, line 1094: Debug code 😅

linter/linter.go
```diff
@@ -606,6 +621,40 @@ func (l *Linter) lintSubRoutineDeclaration(decl *ast.SubroutineDeclaration, ctx
...
+		if gd.IsUsed {
+			l.Error(DuplicatedUseForGotoDestination(stmt.GetMeta(), stmt.Name.Value))
+			return types.NullType
+		}
+
+		gd.IsUsed = true
+		return types.GotoType
+	} else {
+		l.Error(UndefinedGotoDestination(stmt.GetMeta(), stmt.Name.Value))
+	}
```
ysugimoto, line 653: (If I misunderstood, let me know)
The `goto` statement should have a scope conception, otherwise,

```
sub some_function {
  goto foo;
}

sub another_function {
  foo:
}
```

The above code may also be valid.
Hence we need to think about `goto` statement is only available in the subroutine scope, and we should lint it correctly.

ysugimoto (changes requested):
Basically, nice work!
I put some comments, please confirm them 👍

linter/linter.go
shadialtarsha: Actually, you are totally right here! 
I missed this case even tho I have thought about it! 

Thanks for pointing it out, I will fix it

ysugimoto (approved):

PR #73, opened 2022-06-07
Add support for function calls as statements

shadialtarsha:
This PR allows functions to be called without the need for the value to be assigned. Functions like `std.collect` and `resp.tarpit` doesn't need to return a value and can be called as naked statements. 
I am not sure how clean the implementation is, would like to hear your opinion @ysugimoto


ysugimoto (commented):
I understand your thought, the void function calls as a function statement.
I agree with you, but please let me give a few days to review, thanks!

linter/linter.go
```diff
@@ -1271,6 +1273,90 @@ func (l *Linter) lintIfExpression(exp *ast.IfExpression, ctx *context.Context) t
...
+		if len(exp.Arguments) > 0 {
+			err := &LintError{
+				Severity: ERROR,
+				Token:    exp.GetMeta().Token,
+				Message: fmt.Sprintf(
+					"function %s wants no arguments but provides %d argument",
+					exp.Function.String(), len(exp.Arguments),
+				),
+			}
+			l.Error(err.Match(FUNCTION_ARGUMENTS).Ref(fn.Reference))
```
davinci26, line 1298: should we return `return types.NeverType` in this branch

linter/linter.go
```diff
@@ -1271,6 +1273,90 @@ func (l *Linter) lintIfExpression(exp *ast.IfExpression, ctx *context.Context) t
...
+		case types.RTimeType:
+			// fuzzy type check: some builtin function expects RTIME type,
+			// then actual argument type could be STRING because VCL TIME type could be parsed from STRING.
+			if !expectType(arg, types.RTimeType, types.TimeType, types.StringType) {
+				l.Error(FunctionArgumentTypeMismatch(
+					exp.Function.GetMeta(), exp.Function.String(), i+1, v, arg,
+				).Match(FUNCTION_ARGUMENT_TYPE).Ref(fn.Reference))
+			}
+			continue
+		case types.IPType:
```
davinci26, line 1339: do we need tests for all these cases here?

linter/linter.go
```diff
@@ -1271,6 +1273,90 @@ func (l *Linter) lintIfExpression(exp *ast.IfExpression, ctx *context.Context) t
...
+			// fuzzy type check: some builtin function expects IP type,
+			// then actual argument type could be STRING because VCL TIME type could be parsed from STRING.
```
davinci26, line 1341: copy pasta error, you mean VCL IP

davinci26 (changes requested):
lgmt overall, so comments left and right and also we should probably add 100% branch coverage with tests on `linter_test.go`

parser/statement_parser.go
```diff
@@ -557,3 +564,28 @@ func (p *Parser) parseAnotherIfStatement() (*ast.IfStatement, error) {
...
+
+	if !p.peekTokenIs(token.SEMICOLON) {
+		return nil, errors.WithStack(MissingSemicolon(p.curToken))
+	}
+
+	stmt.Meta.Trailing = p.trailing()
+
+	p.nextToken() // point to SEMICOLON
+
+	fmt.Println(stmt.Function)
```
ysugimoto, line 589: Debug code.

ysugimoto (changes requested):
Basically looks good, and I put some comments to make simply.

parser/statement_parser.go
```diff
@@ -557,3 +564,28 @@ func (p *Parser) parseAnotherIfStatement() (*ast.IfStatement, error) {
...
 	return stmt, nil
 }
+
+func (p *Parser) parseFunctionCall() (*ast.FunctionCallStatement, error) {
+	stmt := &ast.FunctionCallStatement{
+		Meta:     p.curToken,
+		Function: p.parseIdent(),
+	}
+
+	p.nextToken() // point to LEFT_PAREN
```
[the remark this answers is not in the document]
ysugimoto: My comment https://github.com/ysugimoto/falco/pull/73#discussion_r899787444 should solve your review.
It should be checked in the previous `switch` section.

__generator__/builtin.yml
```diff
@@ -647,8 +647,8 @@ std.collect:
   reference: "https://developer.fastly.com/reference/vcl/functions/miscellaneous/std-collect/"
   on: [RECV, HASH, HIT, MISS, PASS, FETCH, ERROR, DELIVER, LOG]
   arguments:
-    - [ID]
-    - [ID, STRING]
+    - [STRING]
+    - [STRING, STRING]
```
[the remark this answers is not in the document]
shadialtarsha: This is so tricky! 
Falco treats any `req.%any.%any` as a string and that makes sense because we are evaluating them eventually!
However, `std.collect` treats them as Identifier for some reason. 
I didn't really invest much time to handle this case! 
I would love to hear more from you here and how you would treat this case!

parser/statement_parser.go
```diff
@@ -99,7 +101,12 @@ func (p *Parser) parseBlockStatement() (*ast.BlockStatement, error) {
 		case token.IF:
 			stmt, err = p.parseIfStatement()
 		default:
-			err = UnexpectedToken(p.peekToken)
+			// Check if the current ident is a function call
+			if p.isFunctionCall() {
```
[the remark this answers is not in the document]
shadialtarsha: yeah, I was a bit cautious about adding a case for `token.IDENT` because of potential side effects. 
I will go for it actually!

linter/linter.go
shadialtarsha: hmm not sure why we need `VCP IP`

ysugimoto: Probably, comment typo on this line 😅

ysugimoto (approved):
Looks good to me 👍

linter/linter.go
```diff
@@ -1446,63 +1446,71 @@ func (l *Linter) lintIfExpression(exp *ast.IfExpression, ctx *context.Context) t
...
-			Message:  "If expression returns differernt type between consequence and alternative",
+			Message:  "If expression returns different type between consequence and alternative",
 		})
 	}
 	return left
 }
 
-func (l *Linter) lintFunctionStatement(exp *ast.FunctionCallStatement, ctx *context.Context) types.Type {
-	fn, err := ctx.GetFunction(exp.Function.Value)
+type functionMeta struct {
```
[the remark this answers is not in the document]
ysugimoto: @davinci26 Good point, I propose that we should create a`linter/function.go` and place it.
Only put `functionMeta` struct would be nice, or function related linting method also nice.

parser/statement_parser.go
```diff
@@ -101,15 +101,15 @@ func (p *Parser) parseBlockStatement() (*ast.BlockStatement, error) {
...
-			stmt, err = p.parseGotoDestination()
-		default:
 			// Check if the current ident is a function call
-			if p.isFunctionCall() {
+			if p.peekTokenIs(token.LEFT_PAREN) {
 				stmt, err = p.parseFunctionCall()
 			} else {
-				err = UnexpectedToken(p.peekToken)
+				// Could be a goto destination
+				stmt, err = p.parseGotoDestination()
```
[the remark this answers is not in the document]
shadialtarsha: Added it!

davinci26 (approved):
LGTM thanks!

ysugimoto (approved):
please fix the golangci-lint problem 😄 
I think it's okay to modify `.golangci.yml` to increate threshold

ysugimoto (approved):
Thanks for the great implementation 👍

PR #74, opened 2022-06-07
Updates types for resp.tarpit

davinci26:
After trying it out with fiddle the documentation provided by fastly is wrong.

Update the code to match the behaviour of fiddle rather than the code.

See https://fiddle.fastlydemo.net/fiddle/c61b0d68


ysugimoto (approved):
Nice catch.

Fastly doc says:

```
VOID resp.tarpit( STRING interval_s [, STRING chunk_size_bytes] )
```

I think the document seems wrong, I'll report to the Fastly.

davinci26:
> I think the document seems wrong, I'll report to the Fastly.

We did that already and it confirmed that their docs are not correct and need to be updated

PR #77, opened 2022-06-24
Improve include statement parsing

davinci26:
* Resolve include staments with ".vcl" suffix. This is correct per Fastly linting so we make falco a bit more
flexible.
* Include statements are not required to have a `;` at the end of
the line by fastly. It seems like it is optional.

### DIFF
--- a/cmd/falco/resolver.go
+++ b/cmd/falco/resolver.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"io/ioutil"
@@ -77,7 +78,6 @@ func (f *FileResolver) getVCL(file string) (*VCL, error) {
 	if _, err := os.Stat(file); err != nil {
 		return nil, err
 	}
-
 	fp, err := os.Open(file)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


davinci26:
@ysugimoto any advice on how to test this?

cmd/falco/resolver.go
```diff
@@ -100,13 +100,17 @@ func (f *FileResolver) MainVCL() (*VCL, error) {
 }
 
 func (f *FileResolver) Resolve(module string) (*VCL, error) {
+	module_path_with_extension := module
```
ysugimoto, line 103: Tiny suggestion: the variable name should be lower-camel case.

```suggestion
	modulePathWithExtension := module
```

ysugimoto (changes requested):
I've put nit comment but looks good!

cmd/falco/resolver.go
davinci26: Yeah I was writing this at the same time as I was writing some python code and my brain got all python mode

davinci26: fixed

davinci26:
@ysugimoto added another commit regarding `;` felt like they all fit together into the `include statement` bucket so I added them in the same PR. WDYT?

(cant share fiddles for this, but I have validated this with a fastly service)

ysugimoto:
Ah, I didn't aware of that. If it is valid on Fastly, I think we should respect it.

ysugimoto (approved):

PR #85, opened 2022-07-16
Add support for subroutines operating in multiple scopes

### LINKED ISSUE #84: Allow functions to be used in multiple subroutines
In falco you need to add `@recv` as a comment on top of the funciton but some subroutines can be used in multiple functions if they only use `req` object for example.

davinci26:
The current way we set annotations only allows for subroutines to be used
in a single vcl state function.

Fastly is allows that and it also leads to copy-pasting functions with different
annotations, which is not a good pattern.

I only implemented them as annotations in the form `@<state_name>` instead of supporting them in the
function name because I think it will be weird to have `sub foo_recv_miss_deliver`

Fixes #84

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,13 +99,13 @@ Following table describes subroutine name and recognizing scope:
 
 ### Annotation
 
-For some reasons, the subroutine name could not be changed. Or you want to use this function in multiple scopes.
+For some reasons, the subroutine name could not be changed. Or you want to use this function in multiple scopes. Multiple scopes
+are declared as comma seperated values.
 
 Then, if you apply a hint of scope on annotation, `falco` also understands scope:
 
 ```vcl
-// @recv
-// @miss
+// @recv, miss
 sub custom_process {
    // subroutine has `recv` annotation, lint with RECV|MISS scope.
    // All variables must be accessible in both RECV and MISS scope.
--- a/README.md
+++ b/README.md
@@ -102,17 +102,19 @@ Following table describes subroutine name and recognizing scope:
 For some reasons, the subroutine name could not be changed. Or you want to use this function in multiple scopes. Multiple scopes
 are declared as comma seperated values.
 
-Then, if you apply a hint of scope on annotation, `falco` also understands scope:
+Then, if you apply a hint of scope on annotation, `falco` also understands scope. There are two ways to define the scope annotation:
+1. `@scope: <scope_name1>, <scope_name2>` this is the newest annotation method and it should be prefered over 2.
+2. `@<scope_name1>, <scope_name2>`, this is used to maintain backwards compatibility and it may be deprecated in the future.
 
 ```vcl
-// @recv, miss
+// @scope: recv, miss
 sub custom_process {
    // subroutine has `recv` annotation, lint with RECV|MISS scope.
    // All variables must be accessible in both RECV and MISS scope.
   ...
 }
 
-// @fetch
+// @fetch, miss
 sub custom_request {
   // subroutine has `fetch` annotation, lint with FETCH scope
   ...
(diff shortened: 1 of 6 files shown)


shadialtarsha (approved):
Looks like a nice implementation to me 🚀

ysugimoto (approved):
The implementation looks so nice, and I would suggest one thing a user also can define multiple scopes by separating commas like `@recv,miss`, what do you think about it?

Additionally, I'd suggest that the scope definitions should start with `@scope:` because we may define more annotation-based controls in the future :-)

davinci26:
>  I would suggest one thing a user also can define multiple scopes by separating commas like @recv,miss, what do you think about it?

I can swap to that instead of multiple lines, that might be cleaner.

> re `@scope:`

We should open an issue. My only concern is that this is a breaking change so I am not sure if there is a way to make it easier either via tooling or in some other way

README.md
```diff
@@ -99,16 +99,22 @@ Following table describes subroutine name and recognizing scope:
...
-For some reasons, the subroutine name could not be changed. Then, if you apply a hint of scope on annotation, `falco` also understands scope:
+For some reasons, the subroutine name could not be changed. Or you want to use this function in multiple scopes. Multiple scopes
+are declared as comma seperated values.
+
+Then, if you apply a hint of scope on annotation, `falco` also understands scope:
 
 ```vcl
-// @recv
-sub custom_process { // subroutine has `recv` annotation, lint with RECV scope
+// @recv, miss
```
ysugimoto, line 108: I mean annotation name like `@scope:` would be nice so I expected change with:

```
// @scope: recv,miss
```

This suggestion means we implement _named-annotation_ for additional features in the future.

ysugimoto (changes requested):
Sorry but my comment might not seem to have accuracy.

README.md
davinci26: I see what you are saying but I am worried that this is not back compat, so it will break people when they update.

ivomurrell: Could both the `@scope:` and single `@recv`/`@fetch` etc annotations both be supported for now? Maybe the latter style can be deprecated and support will be dropped in favour of `@scope` in the future.

davinci26: Yeah I could do it as an addition but it might be confusing. Ideally we could have a v2 ideas issue were we append breaking changes and when enough things are added we can have a `--v2` flag

davinci26:
@ysugimoto any middle ground solution that I could implement to make this pr better and also maintain backwards compatibility?

davinci26:
@ysugimoto any thoughts on the above?

README.md
davinci26: added `@scope` as an addition as well

ivomurrell (approved):
this looks great to me, the compromise between the two annotation conventions is pragmatic, and we've been using these changes when linting our VCL for a few weeks now with no issue!

linter/linter_test.go
```diff
@@ -372,6 +373,86 @@ sub vcl_recv BOOL {
 }`
 		assertError(t, input)
 	})
+
+	t.Run("Subroutines can be reused in multiple vcl state functions", func(t *testing.T) {
+		input := `
+//@recv, log
```
ysugimoto, line 379: Does this PR have backward compatibility? If not so, we should change this test with `@scope: recv, log`.

ysugimoto (approved):
LGTM with a nit, we also follow this PR using `@scope` annotation as well 👍

linter/linter_test.go
davinci26: yeah it also maintains backwards compatibility, so both `@scope: recv, log` are supported and also `@recv, log`

davinci26: We can remove the latter in a v2 version of falco if we start having more breaking changes that we want to make

PR #91, opened 2022-07-28
Fix lint of `log` statement

davinci26:
Fastly doesn't allow literals to be passed to log staments but
falco was ok with it. We fix it and add some tests to solidify
the behavior.

Fiddle: https://fiddle.fastlydemo.net/fiddle/1d772f20

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ generate:
 	cd ./__generator__/ && go generate .
 
 test: generate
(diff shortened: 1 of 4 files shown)


ysugimoto (approved):
Nice catch and fixing this 👍 👍

shadialtarsha (approved):

Issue #83, opened 2022-07-12
[BUG] Parser/Linter error: Falco doesnt support calling sub with `()`


davinci26:
Falco doesn't accept the following syntax:

```
sub foo {
  log "1234";
}

sub vcl_recv {

  #Fastly recv
  call foo();
  set req.backend = httpbin_org;

  return (pass);
}
```

With error:

```
💥  Missing semicolon
in /Users/sotiris.nanopoulos/dev/falco/examples/default01.vcl at line 35, position 8
 30|}
 31|
 32|sub vcl_recv {
 33|
 34|  #Fastly recv
 35|  call foo()
```

But fiddle is ok with it:

https://fiddle.fastlydemo.net/fiddle/2a1d3753

davinci26:
If anyone wants to take so be it, otherwise I might be working on it on friday

davinci26:
fixed by https://github.com/ysugimoto/falco/pull/92

PR #92, opened 2022-07-29
Allow calling functions with parentheses

shadialtarsha:
This PR aims to allow functions to be called with parentheses. 
Here are more details about the current issue: https://github.com/ysugimoto/falco/issues/83


ysugimoto (approved):
Looks nice 👍  

The VCL has a bit sloppy syntax, I think  😅

PR #93, opened 2022-08-15
Combine included VCL modules into main and parse once

### LINKED ISSUE #78: [BUG] Parser/Linter Include statement order resolution can be improved
In the following scenario Falco fails but Fastly succeeds:

```
# depenency1.vcl
sub foo {
   set req.backend = httpbin_org;
}

```

```
# main.vcl

backend httpbin_org {
  .connect_timeout = 1s;
  .dynamic = true;
  .port = "443";
  .host = "httpbin.org";
  .first_byte_timeout = 20s;
  .max_connections = 500;
  .between_bytes_timeout = 20s;
  .share_key = "xei5lohleex3Joh5ie5uy7du";
  .ssl = true;
  .ssl_sni_hostname = "httpbin.org";
  .ssl_cert_hostname = "httpbin.org";
  .ssl_check_cert = always;
  .min_tls_version = "1.2";
  .max_tls_version = "1.2";
}

sub vcl_recv {
   call foo;
}
```

This might be a harder change to resolve but I think the correct approach is instead of linting the VCL separately we should be combining them into a single vcl and then parse/lint that as one. I believe this is the behaviour that fastly would do.

ysugimoto:
Fixes #78 

Resolve include modules that are declared at `include "xxx";` statement, replace the parsed AST into its declaration position, and lint whole codes as a single VCL tree

This implementation still does not use DFS style parsing but it's should be lint properly.

### Example

From #78, the example case of VCLs:

```vcl
# dependency1.vcl
sub foo {
   set req.backend = httpbin_org;
}
```

```vcl
# main.vcl

include "dependency1";

backend httpbin_org {
  .connect_timeout = 1s;
  .dynamic = true;
  .port = "443";
  .host = "httpbin.org";
  .first_byte_timeout = 20s;
  .max_connections = 500;
  .between_bytes_timeout = 20s;
  .share_key = "xei5lohleex3Joh5ie5uy7du";
  .ssl = true;
  .ssl_sni_hostname = "httpbin.org";
  .ssl_cert_hostname = "httpbin.org";
  .ssl_check_cert = always;
  .min_tls_version = "1.2";
  .max_tls_version = "1.2";
}

sub vcl_recv {
   call foo;
}
```

Then resolved VCL should be:

```vcl
sub foo {
   set req.backend = httpbin_org;
}

backend httpbin_org {
  .connect_timeout = 1s;
  .dynamic = true;
  .port = "443";
  .host = "httpbin.org";
  .first_byte_timeout = 20s;
  .max_connections = 500;
  .between_bytes_timeout = 20s;
  .share_key = "xei5lohleex3Joh5ie5uy7du";
  .ssl = true;
  .ssl_sni_hostname = "httpbin.org";
  .ssl_cert_hostname = "httpbin.org";
  .ssl_check_cert = always;
  .min_tls_version = "1.2";
  .max_tls_version = "1.2";
}

sub vcl_recv {
   call foo;
}
```

And linter should not raise any errors.

## Additional Specs

- [x] Remote snippets (EdgeDictionary, ACL) also valid
- [x] Nested inclusion also valid
- [x] VCL error of included module still displays its filename and line

### DIFF
--- a/cmd/falco/runner.go
+++ b/cmd/falco/runner.go
@@ -37,7 +37,7 @@ type RunnerResult struct {
 	Infos    int
 	Warnings int
 	Errors   int
-	Vcl     *plugin.VCL
+	Vcl      *plugin.VCL
 }
 
 type StatsResult struct {
@@ -231,7 +231,7 @@ func (r *Runner) Run() (*RunnerResult, error) {
 		Infos:    r.infos,
 		Warnings: r.warnings,
 		Errors:   r.errors,
-		Vcl:     vcl,
+		Vcl:      vcl,
 	}, nil
 }
 
@@ -259,7 +259,7 @@ func (r *Runner) run(v *VCL, mode RunMode) (*plugin.VCL, error) {
 	lt.Lint(vcl, r.context)
 
 	// If runner is running as stat mode, prevent to output lint result
-	if mode & RunModeStat > 0 {
+	if mode&RunModeStat > 0 {
 		return nil, nil
 	}
 
@@ -275,7 +275,7 @@ func (r *Runner) run(v *VCL, mode RunMode) (*plugin.VCL, error) {
 
 	return &plugin.VCL{
 		File: v.Name,
-		AST: vcl,
+		AST:  vcl,
 	}, nil
 }
 
--- a/cmd/falco/runner.go
+++ b/cmd/falco/runner.go
@@ -289,19 +289,11 @@ func (r *Runner) resolveStatements(statements []ast.Statement) ([]ast.Statement,
 			if err != nil {
 				return nil, err
 			}
-			lx := lexer.NewFromString(module.Data, lexer.WithFile(module.Name))
-			p := parser.New(lx)
-			vcl, err := p.ParseVCL()
+
+			vcl, err := r.parseVCL(module.Name, module.Data)
 			if err != nil {
-				lx.NewLine()
-				pe, ok := errors.Cause(err).(*parser.ParseError)
-				if ok {
-					r.printParseError(lx, pe)
-				}
-				return nil, ErrParser
+				return nil, err
 			}
-			lx.NewLine()
-			r.lexers[module.Name] = lx
 
 			vcl.Statements, err = r.resolveStatements(vcl.Statements)
 			if err != nil {
--- a/cmd/falco/runner.go
+++ b/cmd/falco/runner.go
@@ -233,31 +233,39 @@ func (r *Runner) run(v *VCL, mode *RunMode) ([]*plugin.VCL, error) {
 	lx.NewLine()
 	r.lexers[v.Name] = lx
 
-	var vcls []*plugin.VCL
-	// Lint dependent VCLs before execute main VCL
-	for _, stmt := range vcl.Statements {
-		include, ok := stmt.(*ast.IncludeStatement)
-		if !ok {
-			continue
-		}
-
-		if module, err := r.resolver.Resolve(include.Module.Value); err == nil {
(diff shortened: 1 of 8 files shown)


cmd/falco/runner.go
```diff
@@ -51,16 +51,19 @@ type StatsResult struct {
...
-type RunMode struct {
-	isMain bool
-	isStat bool
-}
+type RunMode int
+
+const (
+	RunModeLint RunMode = 0x000001
+	RunModeStat RunMode = 0x000010
+)
```
ysugimoto, line 59: Runner no longer needs to recognize the VCL is main or not, so RunMode could be simple as bit mode.

cmd/falco/runner.go
```diff
@@ -273,11 +269,80 @@ func (r *Runner) run(v *VCL, mode *RunMode) ([]*plugin.VCL, error) {
...
+			vcl, err := r.parseVCL(module.Name, module.Data)
+			if err != nil {
+				return nil, err
+			}
+
+			vcl.Statements, err = r.resolveStatements(vcl.Statements)
+			if err != nil {
+				return nil, err
+			}
+			resolved = append(resolved, vcl.Statements...)
```
ysugimoto, line 302: Resolve `include` statement here.

cmd/falco/runner.go
```diff
@@ -273,11 +269,80 @@ func (r *Runner) run(v *VCL, mode *RunMode) ([]*plugin.VCL, error) {
...
+		case *ast.IfStatement:
+			if err := r.resolveIfStatement(t); err != nil {
+				return nil, err
+			}
+			resolved = append(resolved, t)
```
ysugimoto, line 307: `If` statement may have a block statement so need to resolve recursively.

cmd/falco/runner.go
```diff
@@ -273,11 +269,80 @@ func (r *Runner) run(v *VCL, mode *RunMode) ([]*plugin.VCL, error) {
...
+		case *ast.SubroutineDeclaration:
+			ss, err := r.resolveStatements(t.Block.Statements)
+			if err != nil {
+				return nil, err
+			}
+			t.Block.Statements = ss
+			resolved = append(resolved, t)
```
ysugimoto, line 314: Subroutine may have a block statement so need to resolve recursively.

davinci26:
That is soooooo dope, super excited for this PR. will review tomorrow

ysugimoto:
I'll resolve the conflict lately

cmd/falco/runner.go
davinci26: (also we should probably move this to the code as a comment)

ysugimoto: The block statement may have `include` statement of embedding [dynamic snippets](https://docs.fastly.com/en/guides/using-dynamic-vcl-snippets#including-dynamic-snippets-in-custom-vcl) inside (and probably include file is also valid).

For example:

```vcl
sub vcl_fetch {
  if (req.restarts > 0) {
    include "snippet::<snippet_name>"; # Embed dynamic snippet here
  }
  ...
}
```

According to [include statement](https://developer.fastly.com/reference/vcl/statements/include/), it is available in all subroutines so the above syntax should be valid and resolve inclusion recursively.

ysugimoto:
@davinci26 I've resolved conflict with `main` branch, PTAL?

davinci26 (approved):
LGTM! Thanks for working on this

davinci26:
when this + https://github.com/ysugimoto/falco/pull/99 is merged I will pull it in to our internal branch and test it on our VCL corpus as well and report back if I run into any issues

PR #94, opened 2022-08-17
Allow function expressions in error statements

ivomurrell:
Hey there, I'm currently working on getting the Financial Times' Fastly VCL working with Falco, and have been fixing any false positives I've found. So I might make a few PRs to upstream the fixes over the next few days if that's okay! Thanks for this great resource.

The first change is allowing function calls in `error` statements. We're currently using a `table` to track all the custom error codes we return so that we can minimise duplication and bugs when handling them in `error` subroutines. For example, we might declare an error like
```vcl
error table.lookup_integer(synthetic_errors, "gtg", 601) "Report health on cdn";
```
to (verbosely) enforce the same error code across subroutines (though the non-optional fallback argument doesn't feel great...)

Currently, Falco only expects identifiers or numbers as the first argument to `error`, but functions are also accepted by Fastly. I've added additional parsing logic to allow for this. The linter will also check to make sure that the function returns an integer.

### DIFF
--- a/linter/linter.go
+++ b/linter/linter.go
@@ -1070,6 +1070,11 @@ func (l *Linter) lintErrorStatement(stmt *ast.ErrorStatement, ctx *context.Conte
 		if code != types.IntegerType {
 			l.Error(InvalidType(t.GetMeta(), t.Value, types.IntegerType, code))
 		}
+	case *ast.FunctionCallExpression:
+		code := l.lint(t, ctx)
+		if code != types.IntegerType {
+			l.Error(InvalidType(t.GetMeta(), "error code", types.IntegerType, code))
+		}
 	case *ast.Integer:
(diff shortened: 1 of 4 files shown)


ysugimoto (approved):
Nice implementation, thank you for your PR! Looks so good to me 👍

davinci26 (changes requested):
Thanks for the contribution! We are going through a similar effort at my $company to use falco so we are excited to see what bugs/issues you will uncover.

One small request for some `linter_test.go` tests since we touched the `linter.go` file

davinci26 (approved):
Great thanks for the contribution! Will merge later today unless @ysugimoto beats me to it

ysugimoto:
@davinci26 I've already approved and also looks fine. I'll leave it to you 👍

PR #95, opened 2022-08-18
Fetch backend snippets from Fastly API

ivomurrell:
Falco already supports pulling in some data from the Fastly API with the `--remote` flag but does not currently support getting backends configured in the Fastly UI or via the API. This meant we were seeing a lot of undefined variable errors when linting our VCL, as Falco didn't know about all of our backends.

I've added support for calling the [Backends API](https://developer.fastly.com/reference/api/services/backend/#list-backends) to get these backends now. Additionally, it also checks the `shield` value in each backend, as Fastly will implicitly create a director for each region you have configured to use shields in so that it can balance load across all the shield instances. I've tried to make this latter code a little generic so that it could be refactored and reused if someone wants to add support for fetching the [Director API](https://developer.fastly.com/reference/api/load-balancing/directors/director/#list-directors) later on (though note that this API does not return the shield directors Fastly creates.) The snippet templates are pretty simple as we don't really need any data from the backends/directors we're appending: just their names are required to fix the undefined variable errors really.

### DIFF
--- a/cmd/falco/snippet.go
+++ b/cmd/falco/snippet.go
@@ -37,7 +37,11 @@ backend F_{{ .Name }} {}
 `
 
 var directorTemplate = `
-director {{ .Name }} {{ .Type | printtype }} {}
+director {{ .Name }} {{ .Type | printtype }} {
+	{{- range .Backends }}
+	{ .backend = {{ . }}; .weight = 1; }
+	{{- end }}
+}
 `
 
 type Snippet struct {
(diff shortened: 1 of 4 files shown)


cmd/falco/snippet.go
```diff
@@ -32,6 +32,14 @@ acl {{ .Name }} {
 }
 `
 
+var backendTemplate = `
+backend F_{{ .Name }} {}
+`
+
+var directorTemplate = `
+director {{ .Name }} {{ .Type | printtype }} {}
```
ysugimoto, line 40: falco lints required director properties according to its type (random, fallback, etc..) then name only director may raise a lint error (required properties are declared [here](https://github.com/ysugimoto/falco/blob/main/linter/helper.go#L52)) and fastly docs is [here](https://developer.fastly.com/reference/vcl/declarations/director/).

So I suggest that you can add additional properties to pass the lint.
I assumed from [your implementation](https://github.com/ysugimoto/falco/pull/95/files#diff-a254485a9e67552a5ae5b639419c368b8bbd1d2a03aa36779fb8a2fbdab5c7c9R38-R42) that the shield director type is one of `random`, `client`, or `hash` so the `weight` required property should be declared and some example value like `1` is enough.

ysugimoto (changes requested):
Thanks for adding more remote features!
And I put one request to pass the linting of director 😃

cmd/falco/snippet.go
ysugimoto: Ah sorry I misunderstood, the `.weight` property is required inside the backend definition.
However, falco lints director has at least one backend definition so the director definition should be:

```go
var directorTemplate = `
director {{ .Name }} {{ .Type | printtype }} {
  { .backend = F_{{ .Name }}; .weight = 1; }
}
`
```

Then linter will be passed.

ysugimoto (approved):
Looks good, thanks 👍

PR #99, opened 2022-08-29
Git resolver

### LINKED ISSUE #98: [Feature] Git Based Resolver
The way we use VCL and manage dependencies at $company is by having them versioned in a separate git repository and I would like to build a git file resolver that allows you to resolve paths based on `path:version` and then use `git show $REV:$FILE` to get the file rather than just reading it from disk

So the proposed changes are:

* `GitFileResolver` that implements the `Resolver` interface
* `func NewGitFileResolvers(main string, c *Config) ([]Resolver, error)` and switch between the two implementations based on the command line flag

Addition of two command line flags:

* `-repo              : Path to the git repo that contains the modules`
*  `-git               : Use git to read versioned files that in the form of path/to/file/:version`

WDYT?

davinci26:
Fixes #98 

This adds two new command line flags:

* `-git`: Which controls if paths are treated as versioned paths
* `-repo`: Which set the path to the git repository

Now all include paths are parsed to extract the version (git version) if present. If present we resolve the file with `git show` rather than opening the file directly 

chose to use `git` cmd rather than an in memory git implementation because it is harder with authentication etc


context/context.go
```diff
@@ -163,25 +163,28 @@ func (c *Context) UserDefinedFunctionScope(name string, mode int, returnType typ
...
-// 			set req.http.First = re.group.1;
-// 		}
-// 	}
-// 	set req.http.Third = re.group.2; // difficult to know which (1) or (2) matched result is used
-// }
 //
-// if (req.http.Host) {
-// 	set req.http.Fourth = re.group.3; // difficult to know which (1) or (2) matched result is used or empty
-// }
+//	if (req.http.Host) {
```
davinci26, line 171: linter is not happy without this change 🤷

ysugimoto:
I have a couple of opinions.

1. How about switching the resolver from the target file starting with `git://github.com` ?

Example:

```shell
falco -v git://github.com/org/repo # use GIt resolver
falco -v ./main.vcl # use File resolver
```

2. Why don't we use https://github.com/google/go-github ?

I understand that the `git` command is useful for resolving authentication like private repo, but it means that falco has a dependency on `git` command. Normally `git` command has been installed on almost the developer's machine but we should discuss it.

davinci26:
> Why don't we use https://github.com/google/go-github ?

I started implementing this with https://github.com/go-git/go-git but I thought in the end that it wasnt worth it. Mainly for two reasons:

* Authentication is complicated and it will require us to extend the command line with even more flags.
* We will add a big golang dependency that is used for only a single command

Basically at this point given that we do only one git call it seems like a decent tradeoff not to pull an entire git implementation and rely on a subprocess call. Especially because I didnt have to handle the output and I just take the whole stdout of the process.

> How about switching the resolver from the target file starting with git://github.com ?

Does this mean adding the prefix `git://github.com` for all paths? Or just for the `main.vcl`. That prefix might not be a bad idea since it is used widely

ysugimoto:
> Does this mean adding the prefix git://github.com for all paths? Or just for the main.vcl. That prefix might not be a bad idea since it is used widely

Yes, for example, if 'git://github.com/org/repo` is provided, it adds to include path and resolves include statements from another git repository via `git show` command.

And from another perspective, I think the following commands are the same behavior for managing VCLs on another repository:

```shell
git clone git@github.com:org/repo # ssh or https
falco -I ./repo path/to/main.vcl
```

The above syntax is the common use case and clearly divides the matter of concern git authentication I think, so I'm not sure that the worth of supporting git communication (of course git submodule is not a good way, but...)

It is useful for the continuous integration of local development, Or if there are more benefits, please let me know.

davinci26:
```
git clone git@github.com:org/repo # ssh or https
falco -I ./repo path/to/main.vcl
```

Sadly this is not exactly the same and this is the problem I run into:


Imagine that you have repo 1 that contains:

* `main_1.vcl`
* `main_2vcl`

Then main_1.vcl could depend on version 1 of a dependency and main_2.vcl on version 2. And even within just `main_1.vcl` it could have a dependency that is on `version 1` and another that is on `version 2` so you can checkout the repo, you have to `show` each dependency on the correct version

ysugimoto:
@davinci26 Hmm, could you show me in detail of use case using github repo?

- What does the version mean? fastly's service version?
- Is there a case of managing VCL across versions?

I could not imagine that some VCL file depends on other versions. In my use case, all VCL files indicate one version and it's the standard way of managing VCL.

@shadialtarsha @ivomurrell could you hear me your opinition on this case?

ivomurrell:
If I understand correctly than some special tooling is required when deploying VCL that specifies Git versions for dependencies, as Fastly's [`include`](https://developer.fastly.com/reference/vcl/statements/include/) statement does not support specifying revisions itself. Unless this format of resolving files via a `path:version` string is standardised and used across various organisations I'm not sure whether we should be supporting it in the linter. Is this something that needs to be upstreamed, @davinci26?

ivomurrell:
@davinci26 I can definitely see the utility of the approach you're describing, but the problem still seems to be that the rollout facility depends on custom deployment tooling to parse and resolve the special `path:version` syntax. I'm not convinced it makes sense to put this in the upstream `falco` repository rather than keep as a fork you can maintain when we're talking about adding support for a bespoke solution specific to your company 🙂 Is the code you use to resolve the git revisions when making deployments to Fastly available to look at somewhere?

davinci26:
Realized that the TF mode is solving all my problem and I can TF plan to do this for me. I think this is a win-win overall :) 

This is no longer needed, thanks a lot for the feedback tho!

davinci26:
See https://github.com/ysugimoto/falco/pull/102

PR #100, opened 2022-09-09
Add missing backend property

davinci26:
Property `bypass_local_route_table` is missing and defined in
https://developer.fastly.com/reference/vcl/declarations/backend/


davinci26:
A bit orthogonal to this PR but we should probably have tests that ensure that all the examples compile properly

shadialtarsha (approved):

ysugimoto (approved):

ivomurrell (approved):

PR #102, opened 2022-09-11
Adds support for parsing Dictionary/ACL/Backend from TF json

davinci26:
Some properties can be defined externally in TF and then be used from the VCL file. We add support for parsing them from the TF json file and adding them to the json context.


cmd/falco/terraform.go
```diff
@@ -20,17 +20,37 @@ type TerraformVcl struct {
...
+	Name string
+}
+
+type TerraformDictionary struct {
+	Name string
+}
+
+// TODO(davinci26): We can unmarshall all the properties from the TF file
+// and lint them to make sure they have sane values.
+type TerraformBackend struct {
```
shadialtarsha, line 33: Big opportunity here for us!

shadialtarsha (changes requested):
What would happen if we have double declarations (another declaration in the VCL)
Should we address this case?

cmd/falco/resolver.go
```diff
@@ -22,167 +13,24 @@ type VCL struct {
...
 
-func (s *StdinResolver) Resolve(module string) (*VCL, error) {
-	for i := range s.Modules {
-		if s.Modules[i].Name == module {
-			return s.Modules[i], nil
-		}
-	}
-
-	return nil, errors.New(fmt.Sprintf("Failed to resolve include file: %s.vcl", module))
+// Resolver is an interface for integrate VCL input from file or JSON (terraform planned data)
```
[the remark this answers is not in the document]
davinci26: improved

davinci26:
>What would happen if we have double declarations (another declaration in the VCL) Should we address this case?

This was handled already but I also added a test

davinci26:
See https://reddit.atlassian.net/wiki/spaces/TPORT/pages/2421096466/Developing+deploying+a+Fastly+service and if you have any question reach out to #infra-transport or open a ticket at [go/transport-ticket](https://go.snooguts.net/transport-ticket)

ivomurrell (commented):
Very excited to see this PR: these changes were something that was on my backlog too so that we could get terraform integrated with our own linter workflow 🙌🏻

My main thought is that there seems to be a lot of duplicated logic with the [`cmd/falco/snippet.go`](https://github.com/ysugimoto/falco/blob/main/cmd/falco/snippet.go) code. They're both doing similar things of fetching additional Fastly data, either from the API or from the terraform plan, but they augment the AST in different ways. The snippets add to the VCL source code before parsing whereas this code adds to the AST directly.
I think it would make sense to combine these two behaviours rather than adding more complexity to the `Resolver` struct. As an illustration of why I think this would be of benefit: the `snippets.go` code [supports](https://github.com/ysugimoto/falco/blob/main/cmd/falco/snippet.go#L177) adding shield directors specified in the backends to the AST whereas the `terraform.go` code currently does not. Fastly implicitly creates these directors based on specific backend settings and it can be a bit fiddly to infer them and recreate them. It would be quite fragile to rewrite that logic here and have to maintain it in two places.
I suggest that we preserve the old `Resolver` interface, which can only resolve VCL logic (either from a file or terraform for now), and add a new interface (`Fetcher`?) that can fetch additional data from either terraform or the Fastly API so that we can decouple the `snippets.go` logic from the latter. I think it makes sense to still augment the AST by appending additional source text to it before parsing so that we can support [snippets](https://registry.terraform.io/providers/fastly/fastly/latest/docs/resources/service_vcl#nestedblock--snippet) defined in terraform in the future.

What do you think about this suggestion @davinci26 et al? I'm very happy to help with this refactor if you're happy to go forward with it. Thanks again for all the work done so far!!

ivomurrell:
In retrospect, I should probably have renamed `snippets.go` to something more descriptive in #95 as its utility has grown beyond snippets now...

linter/linter.go
```diff
@@ -54,32 +54,44 @@ func (l *Linter) Lint(node ast.Node, ctx *context.Context) types.Type {
...
 
 func (l *Linter) lintUnusedTables(ctx *context.Context) {
-	for _, t := range ctx.Tables {
+	for key, t := range ctx.Tables {
 		if t.IsUsed {
 			continue
 		}
-		l.Error(UnusedDeclaration(t.Decl.GetMeta(), t.Name, "table").Match(UNUSED_DECLARATION))
+		if t.Decl == nil {
+			l.Error(UnusedExternalDeclaration(key, "backend").Match(UNUSED_DECLARATION))
```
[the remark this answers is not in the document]
davinci26: fixed

davinci26:
@ivomurrell WDYT of the current approach?

cmd/falco/runner.go
```diff
@@ -112,6 +119,16 @@ func NewRunner(rz Resolver, c *Config) (*Runner, error) {
...
+	if f != nil {
+		ctx, timeout := _context.WithTimeout(_context.Background(), 20*time.Second)
+		defer timeout()
+		snippets, err := NewSnippet(f).Fetch(ctx)
+		if err != nil {
+			writeln(red, err.Error())
+		}
+		r.snippets = append(r.snippets, snippets...)
+	}
+
```
shadialtarsha, line 131: If I understand correctly the behavior here. We are trying to load the runner with snippets, backends, ACLS, and Dicts from either remote or Fastly depending on the config.

I think sharing the same Interface between Terraform "Fetcher" and Fastly Fetcher doesn't make sense in my opinion and you can tell from the `context` object that you are passing to the Terraform fetcher without the need for it! 
I don't believe we need to unify the 2 interfaces. we can simply pass them to the runner before. We can do a Fastly call before reaching the `runner` and let the runner takes it from there! 

I don't think the runner should take care of the behavior or filling these objects

Wdyt?

shadialtarsha (changes requested):

cmd/falco/runner.go
davinci26: simplified the interface by taking the context in the constructor via a timeout

linter/linter_test.go
```diff
@@ -1506,6 +1507,23 @@ acl foo {}
...
 		}
 	})
+
+	t.Run("raise unused external error", func(t *testing.T) {
+		input := ` sub foo{}
+`
+		vcl, err := parser.New(lexer.NewFromString(input)).ParseVCL()
+		if err != nil {
+			t.Errorf("unexpected parser error: %s", err)
+			t.FailNow()
```
shadialtarsha, line 1517: A suggestion here would be using `t.Fatalf`

shadialtarsha (approved):
I like the implementation overall
Thanks for adding tests as well! 
I believe right now it is also easy to extend the input to be either yaml or toml or whatever

cmd/falco/snippet.go
```diff
@@ -142,18 +135,20 @@ func (s *Snippet) fetchAccessControl(c _context.Context, version int64) ([]snipp
...
-func (s *Snippet) fetchBackend(c _context.Context, version int64) ([]snippetItem, error) {
-	backends, err := s.client.ListBackends(c, version)
+func (s *Snippet) fetchBackend() ([]snippetItem, error) {
+	var snippets []snippetItem
+	backends, err := s.fetcher.Backends()
 	if err != nil {
-		return nil, fmt.Errorf("failed to get backends for version %w", err)
+		return nil, fmt.Errorf("Failed to get Backends: %w", err)
+	}
+	if len(backends) == 0 {
```
ivomurrell, line 144: **question**: was the Fastly API remote code failing previously if there weren't any backends in your Fastly service, or is this an optimisation?

terraform/fetcher.go
```diff
@@ -0,0 +1,51 @@
...
+		services: s,
+	}
+}
+
+func (f *TerraformFetcher) Backends() ([]*remote.Backend, error) {
+	var b []*remote.Backend
+	for _, s := range f.services {
+		for _, serviceBackend := range s.Backends {
+			b = append(b, &remote.Backend{
+				Name: serviceBackend.Name,
```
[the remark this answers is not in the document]
davinci26: +1 will do

cmd/falco/runner.go
```diff
@@ -51,6 +51,12 @@ type StatsResult struct {
 	Lines       int    `json:"lines"`
 }
 
+type Fetcher interface {
+	Backends() ([]*remote.Backend, error)
```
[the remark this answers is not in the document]
davinci26: I thought about this as well and I went for the lazy way since I wouldn't do anything besides reusing the same abstraction pretty much but long term we should have our own internal abstraction rather than reusing what fastly provides.

Will change that

cmd/falco/snippet.go
davinci26: it failed with if there are no backends so I resolved it by returning early

ivomurrell: ah thanks for fixing this then, sorry I missed it in my PR!

cmd/falco/runner.go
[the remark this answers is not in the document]
davinci26: fixed in 69679c7bb5276feb8b561e2faa2af3b69794b4c1

terraform/fetcher.go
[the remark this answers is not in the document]
davinci26: fixed in 69679c7bb5276feb8b561e2faa2af3b69794b4c1

ivomurrell (approved):
Great stuff, thank you very much!

ysugimoto (approved):
Great works, thank you very much for this!

PR #106, opened 2022-09-27
Sanitize invalid chars in backend name before adding them as snippets

davinci26:
Invalid characters are automatically translated into `_` by fastly on the generated VCL. Match that behaviour in Falco.

### DIFF
--- a/cmd/falco/snippet.go
+++ b/cmd/falco/snippet.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bytes"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"text/template"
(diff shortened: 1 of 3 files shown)


davinci26:
Run into this issue as I was integrating with more services of ours

cmd/falco/snippet.go
```diff
@@ -39,6 +40,12 @@ director {{ .Name }} {{ .Type | printtype }} {
 }
 `
 
+func TerraformBackendNameSanitizer(name string) string {
+	invalid := regexp.MustCompile(`\W`)
```
ysugimoto, line 44: nit: `regexp.MustCompile` may cause panic so I suggest to be placed in package global variables in order to avoid runtime panic.

ysugimoto (approved):
Nice work! I put a nit of golang syntax, but implementation is good!

cmd/falco/snippet.go
ysugimoto: @davinci26 How about this?

It's okay to merge but making regex pointer should be placed on package global.

davinci26: Sorry totally forgot about this :(

davinci26: Will push a commit later today

ysugimoto (approved):
Great work!

PR #107, opened 2022-09-28
Add more implicit conversion fuzy types

davinci26:
According to fastly https://developer.fastly.com/reference/vcl/types/ all primitive types can be casted to string.

> These types all have implicit conversions to strings,
such that their values may be used in contexts where a STRING value is necessary.

We don't allow that in Falco so changed it here to match what Fastly accepts.

### DIFF
--- a/__generator__/builtin.yml
+++ b/__generator__/builtin.yml
@@ -651,8 +651,8 @@ std.collect:
   reference: "https://developer.fastly.com/reference/vcl/functions/miscellaneous/std-collect/"
   on: [RECV, HASH, HIT, MISS, PASS, FETCH, ERROR, DELIVER, LOG]
   arguments:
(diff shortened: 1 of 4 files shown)


shadialtarsha (commented):
Related to this, I had to change the type of parameter that `std.collect` to accept `STRING` instead of `ID`as a workaround to this! 
https://developer.fastly.com/reference/vcl/functions/miscellaneous/std-collect/
Can we change it back to `ID` in `builtin.yml`? or that won't work?

davinci26:
That is a different thing after all so it didnt work

shadialtarsha (approved):
LGTM, nice way to workaround the previous check!

ysugimoto (approved):
Looks good, thanks!

PR #108, opened 2022-09-28
lint protected HTTP headers

ysugimoto:
Fastly protects some HTTP headers modification in VCL so the user could not modify them.
This PR validates the user will modify on `set`, `add`, `unset`, and `remove` statements and raise an error for that operation.

ref: https://developer.fastly.com/reference/http/http-headers/


davinci26 (approved):
nice!

PR #109, opened 2022-09-29
Fix error with `ast.GOTO` and Encoder

davinci26:
Because `ast.GOTO` is not registered in the enccoder the following error appears after the liniting

`Failed to encode VCL: gob: type not registered for interface: ast.GotoStatement`


ysugimoto (approved):
Thanks!

PR #113, opened 2022-10-04
treat obective access in req.http contains semicolon character

ysugimoto:
Related: https://github.com/ysugimoto/falco/pull/112

I found a problem that `req.http.Cookie` allows to contain "." character and then falco raises an "undefined variable" error like:

```vcl
declare local var.cookie STRING;
set var.cookie = req.http.Cookie:dot-included.name
```

Then falco recognizes its syntax as object/variable access and raises an error of "Undefined Variable".
It causes by `splitName` function so I fixed it when ":" character is found in splitted identity names, concatenate remainings.


davinci26 (approved):
thanks, good catch

PR #115, opened 2022-10-19
fix panic error for getting remote snippet

ysugimoto:
This PR avoids unexpected out-of-range panic when `remote` option is provided.

On generating a remote snippet, sometimes nothing backend is defined in Faslty console, then generating a director snippet causes panic by accessing `backends[0]` for empty backends.

The error message is:
```
Remote option supplied. Fetch snippets from Fastly.
Fetching Edge Dictionaries...Done
Fatching Access Control Lists...Done
Fatching Backends...panic: runtime error: index out of range [0] with length 0
```

It causes on https://github.com/ysugimoto/falco/blob/main/cmd/falco/snippet.go#L214

### DIFF
--- a/cmd/falco/snippet.go
+++ b/cmd/falco/snippet.go
@@ -174,7 +174,7 @@ func (s *Snippet) fetchBackend() ([]snippetItem, error) {
 		})
 	}
 
-	// Generate director snippet at least one backend is declared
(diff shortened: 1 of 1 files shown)


davinci26 (approved):
thats great!

PR #121, opened 2023-02-11
fix fourth argument type of accept.language_filter_basic

ysugimoto:
Fix `accept.language_filter_basic` function argument type.
The 4th argument type must be `INTEGER` but defined as `STRING` so I've changed it.

see doc: https://developer.fastly.com/reference/vcl/functions/content-negotiation/accept-language-filter-basic/


ysugimoto:
This is just tiny change so I'll merge this

Issue #110, opened 2022-10-03
[Feature] falco-ignore


kevin-fagan:
Fastly VCL supports some hidden features such as `map` and `callback` that are undocumented. Falco does not support these functions which causes the lint to fail every time. I think it could be beneficial to include some way to tell Falco to ignore the next line, sub-routine, or file.

For example, if you use the `prettier` linter for JS, you can tell prettier to ignore the next line like such:

```js
// prettier-ignore
let positiveInts = test.map((value) => {
    if (value > 0) {
        return value
    }
})
```

We could potentially do the same with Falco:

```VCL
// falco-ignore
map(var.test, test_callback, ",");
```

This can also act as a bandaid to any unexpected bugs a developer might run into while using Falco. If a developer is experiencing some other linting issues that is causing their lint to fail, they can always tell Falco to skip the next like, sub-routine or file to resolve said issues until the linting bug can be patched.

ysugimoto:
Yes, long-time I'm thinking about it.
We'll try to implement that is able to ignore or suppress the error reporting mechanism.

davinci26:
huge +1 on this one

shadialtarsha:
I would like to discuss a proposal for implementing this feature with you here and then I can take it to implementation.

**Feature's Acceptance Criteria** 
1. 2 levels for ignoring will be introduced:
- On the subroutine level
- On the statement Level
2. This feature shouldn't interfere with the current semantic comments that falco has. To be more specific: Fastly macro comments and Scope comments 
3. Tokens should be parsed even with ignore in place. Ignoring's effect should only be during the linting phase.
4. (optional) Show what linting errors being ignored

**High-level implementation**
Introduce 2 new comments:
- `--falco-ignore`: Don't lint the next statement!
- `--faloc-sub-ignore`: Don't lint the next subroutine block

Inside the declaration of the subroutine and the statement, we will add a new field that indicates if any of them have these comments set! - Or we can use the `infixComment` - but I need to double check the impelementation of this comment.

With these two new comments, we introduce two new Flags to `Context`: 
- `falcoIgnore: false`
- `falcoSubIgnore: false`
setting and unsetting the values for these 2 flags will happen in 2 places: 
- `lintSubroutineDeclaration`
- `lintBlockStatement`

Suppressing the errors will happen based on the value of these flags.

So what do we think here? I would love to hear your feedback and your ideas!

kevin-fagan:
[...]

PR #140, opened 2023-04-20
implement ignoring feature

ysugimoto:
Fixes https://github.com/ysugimoto/falco/issues/110

This PR implements the ignoring feature by appending/prepending specific comments.
Regarding https://github.com/ysugimoto/falco/issues/110 , this is partial support but this PR is enough I think for now.
I implemented following signatures:

- `// falco-ignore-next-line` on leading comment
- `// falco-ignore` on trailing comment
- `// falco-ignore-start` and `// falco-ignore-end` on leading comment

And, we does not support ignoring specific rules like `// falco-ignore unused/variable` for now.
This should be implemented when we need but we can do it above this implementation. This PR is underlying.

Here are examples:

### Ignore any errors whole subroutine

```vcl
// falco-ignore-start
sub vcl_recv {
  ...  any statement/expression errors will be ingored
}
// falco-ignore-end
...
```

### Ignore any errors for the next statement

```vcl
sub vcl_recv {
  // falco-ignore-next-line
  set req.http.X-Value = some.undefined.value;
}
```

### Ignore any errors in the current statement

```vcl
sub vcl_recv {
  set req.http.X-Value = some.undefined.value; // falco-ignore
}
```

If anyone have opinion, let's discuss for that.

### DIFF
--- a/linter/ignore.go
+++ b/linter/ignore.go
@@ -1,0 +1,100 @@
+package linter
+
+import (
+	"strings"
+
+	"github.com/ysugimoto/falco/ast"
+)
+
+// ignore signatures
+const (
+	falcoIgnoreNextLine = "falco-ignore-next-line"
+	falcoIgnoreThisLine = "falco-ignore"
+	falcoIgnoreStart    = "falco-ignore-start"
+	falcoIgnoreEnd      = "falco-ignore-end"
+)
+
+type ignore struct {
+	ignoreNextLine bool
+	ignoreThisLine bool
+	ignoreRange    bool
+}
+
+// Setup ignores for common statements, declarations.
+// For common statements:
+//
+// // leading comments
+// [STATEMENT] // trailing comments
+//
+// Then leading comments accept falco-ignore-next-line, falco-ignore-start, falco-ignore-end
+// trailing comments accept falco-ignore
+func (i *ignore) SetupStatement(meta *ast.Meta) {
+
+	// Find ignore signature in leading comments
+	for _, c := range meta.Leading {
+		line := strings.TrimLeft(c.String(), "#@*/ ")
+		switch {
+		case strings.HasPrefix(line, falcoIgnoreNextLine):
+			i.ignoreNextLine = true
+		case strings.HasPrefix(line, falcoIgnoreStart):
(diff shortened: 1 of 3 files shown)


davinci26:
Looks awesome, will review today. Super exciting!

linter/linter.go
```diff
@@ -268,7 +272,11 @@ func (l *Linter) lintVCL(vcl *ast.VCL, ctx *context.Context) types.Type {
 
 	// Lint each statement/declaration logics
 	for _, s := range statements {
-		l.lint(s, ctx)
+		func(v ast.Statement, c *context.Context) {
```
davinci26, line 275: should we define this as a helper function?

linter/ignore.go
```diff
@@ -0,0 +1,98 @@
...
+			i.ignoreNextLine = true
+		case strings.HasPrefix(line, falcoIgnoreStart):
+			i.ignoreRange = true
+		case strings.HasPrefix(line, falcoIgnoreEnd):
+			i.ignoreRange = false
+		}
+	}
+
+	// Find ignore signature in trailing comments
+	for _, c := range meta.Trailing {
```
davinci26, line 46: What would happen in this case?

```
set req.http.foo = "bar"; // falco-ignore-start
set req.http.bar = underfined
//falco-ignore-end
```
The `falco-ignore-start` would be ignored right?

davinci26 (approved):
pretty clean implementation, one small comment and one question.

linter/linter.go
ysugimoto: You are right, will do 👍

linter/ignore.go
ysugimoto: Yes. the `// falco-ignore-start` expects that present in the leasing comment.
It means the trailing comment only accepts `// falco-ignore`, which is related to the current statement.

This comes from our parser logic:

```
set req.http.foo = "bar"; // falco-ignore-start  <- trailing comment
// falco-ignore-start <- leading comment
set req.http.foo = "bar";
```

So linefeed is important.

ysugimoto:
I'll merge this and release with minor version

Issue #147, opened 2023-05-08
[BUG] Terraform Resolver Module's Look up is Failing


shadialtarsha:
**Describe the problem**
The Terraform Resolver is not able to look up the modules correctly.

**VCL code that causes the problem**
We are trimming the extension [here](https://github.com/ysugimoto/falco/blob/6fcd3db06970550b53eeccb1d6aaf12668d7b747/resolver/terraform.go#L52) and then we try to look up the module name. However, the modules' names have the `vcl` extension.
Removing this line will fix the issue!

I will issue a PR but I just want to check if I am missing something here!

ysugimoto:
Resolver seems fine even if we don't trim the extension I think.
I look forward to seeing your PR!

ysugimoto:
fixed on https://github.com/ysugimoto/falco/pull/148

PR #148, opened 2023-05-10
Fix resolving Terraform Modules

shadialtarsha:
A fix for https://github.com/ysugimoto/falco/issues/147

### DIFF
--- a/cmd/falco/runner_test.go
+++ b/cmd/falco/runner_test.go
@@ -88,6 +88,25 @@ func TestResolveWithDuplicateDeclarations(t *testing.T) {
 	}
 }
 
+func TestResolveWithMultipleModules(t *testing.T) {
+	fileName := "../../terraform/data/terraform-modules.json"
+	rslv, f := loadFromTfJson(fileName, t)
+
+	r, err := NewRunner(&Config{V: true}, f)
+	if err != nil {
+		t.Fatalf("Unexpected runner creation error: %s", err)
+	}
+
+	ret, err := r.Run(rslv[0])
+	if err != nil {
+		t.Fatalf("Unexpected Run() error: %s", err)
+	}
+
+	if ret.Errors > 0 {
+		t.Errorf("Errors expects 0, got %d", ret.Errors)
+	}
+}
+
 // Adds a test for all the example code in the repo to make sure we don't accidentally
 // break those as they are the first thing someone might try on the repo.
 func TestRepositoryExamples(t *testing.T) {
--- a/resolver/terraform.go
+++ b/resolver/terraform.go
@@ -2,7 +2,6 @@ package resolver
 
 import (
 	"fmt"
-	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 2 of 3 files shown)


cmd/falco/runner_test.go
```diff
@@ -88,6 +88,25 @@ func TestResolveWithDuplicateDeclarations(t *testing.T) {
...
+	ret, err := r.Run(rslv[0])
+	if err != nil {
+		t.Fatalf("Unexpected Run() error: %s", err)
+	}
+
+	if ret.Errors > 0 {
+		t.Errorf("Errors expects 0, got %d", ret.Errors)
+	}
+}
+
```
ysugimoto, line 109: Could you add more test cases to include module without `.vcl` extension please?

ysugimoto (commented):
Basically good, and then I added one suggestion about backward compatibility.

resolver/terraform.go
```diff
@@ -23,6 +23,11 @@ func NewTerraformResolver(services []*terraform.FastlyService) []Resolver {
 			ServiceName: v.Name,
 		}
 		for _, vcl := range v.Vcls {
+			// Always save module names with .vcl extension
+			if !strings.HasSuffix(vcl.Name, ".vcl") {
+				vcl.Name = fmt.Sprintf("%s.vcl", vcl.Name)
+			}
```
shadialtarsha, line 29: @ysugimoto I think we should always assume the statement `include "filename.vcl"` should has an extension no matter what, thus we should always insert the module names with extensions. what do you think?

ysugimoto: My concern is that trims `.vcl` extension on resolving module.
By removing the extension but even `stmt.Module.Value` may have `some_module.vcl` then could we solve it?

I means that a custom vcl is written something like `include "some_module";` I'm guessing it may not solve them so I just want to regard to do it by adding test cases.

shadialtarsha: ahh so you think it is ok to have `include "some_module"` and not `include "some_module.vcl"`
I thought the include statement should be followed by "module_name + vcl" always

shadialtarsha: If you think we should allow this case, I can add that!

ysugimoto: Yes I think we should allow that case because my project defines custom vcl without extension.
For example,

```vcl
// main.vcl
include "some_module";
```

```hcl
// main.tf
resource "fastly_service_vcl" "..." {
... 
  vcl {
    name = "some_module"
    content = "..."
  }
...
```

So I'm happy to allow the above case for backward compatibility!

shadialtarsha:
@ysugimoto I decided to always have them internally with `.vcl`! 
We can do it also the other way around, and have them without an extension! It is up to you

shadialtarsha:
@ysugimoto can I get another round of review please?

ysugimoto (approved):
Looks good at all 👍

ysugimoto (approved):

PR #149, opened 2023-05-25
Fixup: Accept `.backend` in chash director

davinci26:
Previously it wasn't accepted and this is not correct based on fastly [examples](https://developer.fastly.com/reference/vcl/declarations/director/#consistent-hashing)


ysugimoto (approved):
Looks good, nice work!

PR #150, opened 2023-05-26
add lint rule of allow-public-url-purge

ysugimoto:
Add new linting rule of `allow-public-url-purge` .

This rule means the URL purge request should be authenticated via API, not be called publicly,
because it causes unexpected cache purging from anonymous users and the traffic will pass to the origin.

ref: https://docs.fastly.com/en/guides/authenticating-api-purge-requests

Of course, this is a matter of the service policy so this rule's severity should be `INFO`.


ysugimoto:
This linting rule could not run correctly on linter so I'm going to close this PR

PR #151, opened 2023-05-30
integrate configuration to config package

ysugimoto:
Now we have some configurations to control command behavior -- cli options, `.falcorc` file, and environment variables.
Then this PR integrates these configurations into a struct of `config.Config`.

On the command has started, the process integrates configurations and we can use all options through the config struct.
This struct automatically cascades configurations following order:

1. find `.falco.yaml` and parse it (Note that the configuration filename has changed from `.falcorc` to `.falco.yaml`
2. parse command-line options
3. map environment variables

To integrate, we have new dependencies of https://github.com/ysugimoto/twist
For future improvement, the integrated config struct makes us useful for adding more options, configuration files, etc...


ysugimoto:
I will merge this PR. If you have any suggestion, let me know in another PR.

PR #155, opened 2023-07-21
Add fastly_info.host_header

bungoume:
Support for new variables fastly_info.host_header listed in the documentation.
https://developer.fastly.com/reference/vcl/variables/client-request/fastly-info-host-header/


ysugimoto (approved):
LGTM

Issue #156, opened 2023-08-19
[Feature] JSON support for generic, terraform and lint usage


doramatadora:
**Describe the feature you'd like**
When `falco` is run without a subcommand, or with a subcommand other than `stat`, it should be possible to obtain JSON results for the run.

**Put example VCL after the feature has been implemented**
```
sub vcl_fetch {

  #Fastly fetch
  return(deliver)
}
```
Running `falco -json ./myVclFile.vcl` should result in:
```json
{
  "Infos": 0,
  "Warnings": 0,
  "Errors": 0,
  "LintErrors": {},
  "ParseErrors": {
    "/Users/dmilitaru/fsly/vscode-vcl/myVclFile.vcl": {
      "Token": {
        "Type": "RIGHT_PAREN",
        "Literal": ")",
        "Line": 4,
        "Position": 17,
        "Offset": 0,
        "File": "/Users/dmilitaru/fsly/vscode-vcl/myVclFile.vcl",
        "Snippet": false
      },
      "Message": "Missing semicolon"
    }
  },
  "Vcl": null
}
```

Running `falco -vv -json ./falco/examples/default03.vcl` should result in:
```json
{
  "Infos": 1,
  "Warnings": 0,
  "Errors": 0,
  "LintErrors": {
    "/Users/dmilitaru/fsly/vscode-vcl/falco/examples/default03.vcl": [
      {
        "Severity": "Info",
        "Token": {
          "Type": "REGEX",
          "Literal": "~",
          "Line": 34,
          "Position": 23,
          "Offset": 0,
          "File": "/Users/dmilitaru/fsly/vscode-vcl/falco/examples/default03.vcl",
          "Snippet": false
        },
        "Message": "regex captured variable may override older one",
        "Reference": "",
        "Rule": "regex/matched-value-override"
      }
    ]
  },
  "ParseErrors": {},
  "Vcl": {
      // PARSED VCL AST 
   }
}
```

**Additional context**
`falco` is amazing! I would like to use it in a Language Server I'm building for Fastly VCL.

ysugimoto:
Hi @doramatadora, thanks for the suggestion!

> I would like to use it in a Language Server I'm building for Fastly VCL.

I'm excited about your thought! So, what do you think that we make a new subcommand `diagnosis` for the language server?
Then you can use `falco diagnosis ./falco/examples/default03.vcl` for the specific way and you can get JSON as you wrote:

```json
{
  "Infos": 1,
  "Warnings": 0,
[...]

PR #157, opened 2023-08-19
Adds JSON support for generic, terraform and lint usage

doramatadora:
Proposed implementation for https://github.com/ysugimoto/falco/issues/156

This PR proposes that the following commands will emit JSON-serialized output:

- `falco -json`
- `falco terraform -json`
- `falco stat -json`
- `falco terraform -json`

The other main change with this flag is that `ParseError`s aren't surfaced, but marshalled into the emitted JSON.

`-json` ignores `-v` and `-vv`: all linter errors, warnings and suggestions are returned.

### DIFF
--- a/cmd/falco/main.go
+++ b/cmd/falco/main.go
@@ -63,9 +63,9 @@ Flags:
     -h, --help         : Show this help
     -r, --remote       : Communicate with Fastly API
     -V, --version      : Display build version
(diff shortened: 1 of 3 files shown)


ysugimoto (approved):
There are many conditional branches to suppress write to stdout but it should be improved on the other PRs (I will do it)

Thanks for the good implementation!

Issue #185, opened 2023-09-20
[BUG] falco will crush with the fastly service without VCL in the terraform planned output


Shinyaigeek:
**Describe the problem**

falco will crush if falco get the fastly service without VCL in the terraform planned output. I have the project which have the multiple fastly services managed in the single terraform state. These fastly services include the service which does not have VCL. 

I tried to introduce falco linter/tester into this project, but when I run `terraform show -json planned.out | falco terraform` command, falco will crush with the below error output.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x1008da520]

goroutine 1 [running]:
main.(*Runner).run(0x140002ae690, 0x2?, 0x0, 0x1)
        /home/runner/work/falco/falco/cmd/falco/runner.go:227 +0x30
main.(*Runner).Run(0x140002ae690, {0x100bf16c0?, 0x140007b26c0})
        /home/runner/work/falco/falco/cmd/falco/runner.go:211 +0x180
main.runLint(0x140002ae690, {0x100bf16c0?, 0x140007b26c0?})
        /home/runner/work/falco/falco/cmd/falco/main.go:145 +0x30
main.main()
        /home/runner/work/falco/falco/cmd/falco/main.go:131 +0x84c
```

falco's linter/tester/simulator expect all fastly service to have VCL (ref: https://github.com/ysugimoto/falco/blob/main/resolver/terraform.go#L29), so falco will crush when falco encounter the fastly service which does not have VCL. I think falco should filter the fastly services which does not have VCL in advance.

**VCL code that cause the problem**

This is not related to VCL, but I made simple repro; https://github.com/Shinyaigeek/falco-with-multiple-services

**Desktop (please complete the following information):**
 - OS: MacOS arm64
 - Shell zsh

Shinyaigeek:
I send the PR to fix this issue in #186

ysugimoto:
This issue won't be tackled.

PR #186, opened 2023-09-20
Filter the fastly service without VCL in constructing service from the planned output by terraform

Shinyaigeek:
## Summary

Filter the fastly service without VCL in constructing service from the planned output by terraform

Fixes: #185 

I made `UnmarshalTerraformPlannedInput`, which constructs fastly service from terraform planned output, filter the fastly service without VCL. As #185 , falco will crush to handle the fastly service without VCL, so filtering in advance is needed.

If there is a better approach, please feel free to close this PR 👌

### DIFF
--- a/terraform/terraform.go
+++ b/terraform/terraform.go
@@ -136,5 +136,5 @@ func UnmarshalTerraformPlannedInput(buf []byte) ([]*FastlyService, error) {
 
 func isFastlyVCLServiceResource(r *TerraformPlannedResource) bool {
 	return r.ProviderName == fastlyTerraformProviderName &&
-		(r.Type == fastlyVCLServiceType || r.Type == fastlyVCLServiceTypeV1)
+		(r.Type == fastlyVCLServiceType || r.Type == fastlyVCLServiceTypeV1) && len(r.Values.Vcl) > 0
 }
--- a/terraform/data/terraform-without-vcl.json
+++ b/terraform/data/terraform-without-vcl.json
@@ -1,0 +1,114 @@
+{
+    "planned_values": {
+        "root_module": {
+            "resources": [
+                {
+                    "provider_name": "registry.terraform.io/fastly/fastly",
+                    "type": "fastly_service_vcl",
+                    "values": {
+                        "acl": [
+                            {
+                                "acl_id": "this is another id",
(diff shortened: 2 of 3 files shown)


terraform/terraform.go
```diff
@@ -136,5 +136,5 @@ func UnmarshalTerraformPlannedInput(buf []byte) ([]*FastlyService, error) {
 
 func isFastlyVCLServiceResource(r *TerraformPlannedResource) bool {
 	return r.ProviderName == fastlyTerraformProviderName &&
-		(r.Type == fastlyVCLServiceType || r.Type == fastlyVCLServiceTypeV1)
+		(r.Type == fastlyVCLServiceType || r.Type == fastlyVCLServiceTypeV1) && len(r.Values.Vcl) > 0
```
ysugimoto, line 139: Unmarshalled resources should be a Fastly service resource even if VCL is empty.
I would suggest that we should notify the user to skip linting if VCL is empty explicitly.
So we can check VCL is empty [here](https://github.com/ysugimoto/falco/blob/main/cmd/falco/main.go#L127) and skip linting.

ysugimoto (changes requested):
Basically nice but I would suggest one thing to be useful for the user.

terraform/terraform.go
Shinyaigeek: Your proposal makes sense, thank you!

I have one point to discuss about how should we log the services filtered. [In the current implementation](https://github.com/ysugimoto/falco/blob/develop/cmd/falco/main.go#L119-L132), falco runs many commands other than `lint` , such as `test` and `simulate`. Only `lint` command logs which service `falco` is linting. I think there are two ways to log the services filtered;

- First, logging `{command} service of “{service name}”` also in the other commands.
- Second, filtering and logging the above like `{service of “{service name”} is filtered because that does not have VCL` in the line https://github.com/ysugimoto/falco/blob/develop/cmd/falco/main.go#L118

I feel the latter is better, but how about you?

ysugimoto: Sorry for the late, I also think the latter should be better, I agree with you.

ysugimoto:
Any progress? If not, we will close this PR temporarily. Feel free to reopen.

PR #200, opened 2023-11-11
Implement STRING to ACL regex comparison

MasonM:
First off, thank you for creating this tool! It's very impressive, and fulfills a pressing need for my team.

This is the first of hopefully several PRs that I'm planning on entering. This one is simple: it implements comparison between a string and an ACL, re-using the code for IP to ACL comparisons. Before this, VCL like the following triggers the error `Invalid type comparison STRING and ACL` (taken from https://developer.fastly.com/reference/vcl/declarations/acl/):
```
  if (req.http.Fastly-Client-IP ~ ban_acl) {
    error 601 "banned-ip";
  }
```


ysugimoto (approved):
@MasonM
Nice catch, great stuff!
I realize that acl matching is an important feature, really we need this!

PR #201, opened 2023-11-11
Fix type of client.geo.ip_override

MasonM:
`client.geo.ip_override` is of type STRING per https://developer.fastly.com/reference/vcl/variables/geolocation/client-geo-ip-override/, not boolean. I also added some basic tests for this.


ysugimoto (approved):
@MasonM LGTM, thanks! I'll merge this 👍

PR #206, opened 2023-11-13
Allow assigning a backend to a string

MasonM:
The following code:
```
set req.backend = httpbin_org;
declare local var.backend STRING;
set var.backend = req.backend;
```

will trigger the error `Failed to assign value to var.backend, BACKEND identifier could not assign to STRING` when testing/simulating, but this is allowed by Fastly, and is a common workaround to allow doing regex matching on a backend.

### DIFF
--- a/interpreter/assign/assign.go
+++ b/interpreter/assign/assign.go
@@ -134,9 +134,6 @@ func Assign(left, right value.Value) error {
 			rv := value.Unwrap[*value.Time](right)
 			lv.Value = rv.Value.Format(http.TimeFormat)
 		case value.BackendType: // STRING = BACKEND
-			if right.IsLiteral() {
-				return errors.WithStack(fmt.Errorf("BACKEND identifier could not assign to STRING"))
-			}
 			rv := value.Unwrap[*value.Backend](right)
 			lv.Value = rv.Value.Name.Value
 		case value.BooleanType: // STRING = BOOL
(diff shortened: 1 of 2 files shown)


interpreter/assign/assign.go
```diff
@@ -134,9 +134,6 @@ func Assign(left, right value.Value) error {
 			rv := value.Unwrap[*value.Time](right)
 			lv.Value = rv.Value.Format(http.TimeFormat)
 		case value.BackendType: // STRING = BACKEND
-			if right.IsLiteral() {
-				return errors.WithStack(fmt.Errorf("BACKEND identifier could not assign to STRING"))
-			}
```
ysugimoto, line 139: The literal check of the right expression is still needed. `IsLiteral()` method returns bool that the right expression is a literal or not.

This case checks BACKEND literal assignment for STRING variable the following example:

```vcl
set req.backend = F_origin_0;
declare local var.backend STRING;
[VALID] set var.backend = req.backend; // req.backed is not a literal, should do type conversion
[INVALID] set var.backend = F_origin_0; // F_origin_0 is a literal
```

The above code is invalid in Fastly. I put a fiddle here: https://fiddle.fastly.dev/fiddle/108043ab

ysugimoto (commented):
Thanks, but I think your suggestion may not be needed.

MasonM:
@ysugimoto You're right, this fix wasn't correct. I updated the PR to revert the changes and instead add two integration tests  to `interpreter/expression_test.go`: the first checks you can assign `req.backend` to a variable and the second checks that you can't assign a backend literal to a variable. I'm not sure what the right fix is here (I thought at first I could update `Assign()` to set `lv.Literal = true`, but that causes the second test case to fail), but I can look into this later.

MasonM (commented):
Okay, I have a fix that should work without any unwanted side effects. The root problem is that calling `Get(s, "req.backend")` returns the backend with `Literal=true` set, which is wrong, because it's not a literal anymore if it's being accessed through a variable. I updated `Get()` to instead return a copy of the backend, without `Literal`.

I also added a check to ensure a backend literal can't be assigned to a variable (this was an existing issue, not one introduced by the fix) and a bunch of test cases.

PR #207, opened 2023-11-13
Fix testing.override_host

MasonM:
Currently, trying to use `testing.override_host()` in a test results in the following error:
```
$ falco test -I . default.vcl
Running tests... Done.
 FAIL  /Users/mmalone/src/go/src/github.com/ysugimoto/falco/examples/testing/default.test.vcl
  ●  [RECV]  Foo request header should contains "hoge"

    [RuntimeException] Function testing.override_host is not defined in/Users/mmalone/src/go/src/github.com/ysugimoto/falco/examples/testing/default.test.vcl at line: 5, position: 3
```

I tried writing a unit test for the `tester.Run()` method, but it's tricky due to the global variables `injectedVariable` and `builtinFunctions`. Making those part of the `Interpreter` interface would fix that.

### DIFF
--- a/tester/function/testing_functions.go
+++ b/tester/function/testing_functions.go
@@ -50,6 +50,20 @@ func TestingFunctions(i *interpreter.Interpreter, c Counter) map[string]*ifn.Fun
 				return false
 			},
 		},
+		"testing.override_host": {
+			Scope: allScope,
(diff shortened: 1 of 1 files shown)


ysugimoto (approved):
Oh, I forgot to add this section, thanks!

Issue #196, opened 2023-10-20
[BUG] Backend is not found in director (during test) 


mgreenw:
Hi! Thanks for making such a useful tool.

**Describe the problem**
I am seeing an error when running `falco test -r` where something is going wrong with a backend / director related to shielding. When I run `falco test -r`, I get

```
 Failed to run test: [RuntimeException] backend 'F_Custom_backend' is not found inRemote.Director:ssl_shield_region_state_us at line: 3, position: 4
```

When I run `falco lint`, I see a notice related to this:

```
 2|director ssl_shield_region_state_us random {
   ^^^^^^^^
 3|    { .backend = F_Custom_backend; .weight = 1; }
```

**However**, if I look at the generated VCL directly in fastly, I see this, seemingly missing the backend:

```
director ssl_shield_region_state_us shield {
   .shield = "INTER-REGION-SSL";
}
```

**VCL code that cause the problem**
It seems to be that the code is coming from the fastly remote and not from the local vcl code being linted / tested.

**Expected behavior**
I expect shielding / backends to work for linting and testing.

**Desktop (please complete the following information):**
 - OS: macOS 13.3
 - Shell: zsh

ysugimoto:
Hi @mgreenw , thanks for reaching out to us!

This issue looks interesting and it is caused by Origin-Shielding related configuration.

To explain what occurs:

1. On fetching and generating remote snippets, we will add the first backend declaration into `.backend` property to avoid raising an error (see: https://github.com/ysugimoto/falco/blob/main/snippets/fetch.go#L172)
2. In this case, the director looks to have a special type of "shield" (unfortunately it isn't documented in Fastly)
 
So you could see the falco's error such as having a backend and we assume we could fix this problem by ignoring shield director.

We will do that in a few days, thanks again for reporting that issue that we were not aware of!

ysugimoto:
Hmm, I could not reproduce it.

I've created a service that has an origin with origin-shielding, Faslty shows VCL the following:

```vcl
backend F_Host_1 {
    .between_bytes_timeout = 10s;
    .connect_timeout = 1s;
    .dynamic = true;
    .first_byte_timeout = 15s;
    .host = "example.com";
    .max_connections = 200;
    .port = "443";
    .share_key = "qRK2E1vLIVkQ3BU0iVk9X7";
    .ssl = true;
    .ssl_cert_hostname = "example.com";
    .ssl_check_cert = always;
    .ssl_sni_hostname = "example.com";
[...]

PR #208, opened 2023-11-13
Fix backend  not found in director

MasonM:
This should fix https://github.com/ysugimoto/falco/issues/196, or at least one of the causes of these errors. The problem is that `ProcessDeclarations()` method parses the declarations in order, and if a `director` declaration references a `backend` declaration that's defined later, it'll trigger `backend 'F_Custom_backend' is not found`. The fix is to have `ProcessDeclarations()` do two passes: the first to process backends and the second for everything else.

### DIFF
--- a/interpreter/interpreter.go
+++ b/interpreter/interpreter.go
@@ -136,29 +136,37 @@ func (i *Interpreter) ProcessInit(r *http.Request) error {
 }
 
 func (i *Interpreter) ProcessDeclarations(statements []ast.Statement) error {
-	// Process root declarations and statements
+	// Process root declarations and statements.
+	// Must process backends first because they're referenced by directors.
+	// https://developer.fastly.com/reference/vcl/declarations/
 	for _, stmt := range statements {
-		// Call debugger
+		t, ok := stmt.(*ast.BackendDeclaration)
+		if !ok {
+			continue
+		}
 		i.Debugger.Run(stmt)
+		h := &atomic.Bool{}
+		h.Store(true)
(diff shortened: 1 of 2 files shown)


interpreter/interpreter.go
```diff
@@ -136,29 +136,37 @@ func (i *Interpreter) ProcessInit(r *http.Request) error {
...
+		h.Store(true)
+		// Determine default backend
+		if i.ctx.Backend == nil {
+			i.ctx.Backend = &value.Backend{Value: t, Literal: true, Healthy: h}
+		}
+		if _, ok := i.ctx.Backends[t.Name.Value]; ok {
+			return exception.Runtime(&t.Token, "Backend %s is duplicated", t.Name.Value)
+		}
+		i.ctx.Backends[t.Name.Value] = &value.Backend{Value: t, Literal: true, Healthy: h}
+	}
```
ysugimoto, line 158: Suggestion
Could you define this loop as a separated function like `ProcessBackends()` ? it makes be less complex I think.

ysugimoto (changes requested):
Thanks, I realized that this is a backend hoisting problem 👍 
And I put one suggestion about keeping less complexity.

MasonM:
Sure thing. I'm on vacation right now, but I'll do this when I get back on
Friday.

MasonM:
Okay, I pushed a commit to extract the logic to a `ProcessBackends()` function and added some additional tests

ysugimoto (approved):
Looks good, thanks!

PR #209, opened 2023-11-13
Fix Terraform plan parsing when multiple resources present

MasonM:
When defining a Fastly service in Terraform, it's common to reference other resources, such as
[aws_s3_object](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_object). However, these resources often have a different structure for the `"values"` key, which results in the following error when using `falco terraform lint`:
```
 Failed to unmarshal stdin input: json: cannot unmarshal string into Go
 struct field .planned_values.root_module.resources.values.acl of type
 []*terraform.TerraformAcl
```

The fix is to call `json.Unmarshal()` twice: once to identify the location of the Fastly service in the plan, and then again to get the values for that service.

### DIFF
--- a/terraform/data/terraform-valid-with-s3.json
+++ b/terraform/data/terraform-valid-with-s3.json
@@ -1,0 +1,81 @@
+{
+    "planned_values": {
+        "root_module": {
+            "resources": [
+                {
+                    "address": "aws_s3_object.object",
+                    "mode": "managed",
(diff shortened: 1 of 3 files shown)


ysugimoto:
I understand what you intend to fix. good!

ysugimoto (approved):

PR #210, opened 2023-11-13
Fix tester/simulator when shielding used

MasonM:
The changes in 39d728808a527605bf5944b95cd1f70fb224d113 are causing the following error when testing/simulating a service that uses shielding:
```
At least one backend must be specified in director 'ssl_shield_this_is_a_shield' in Remote.Director:ssl_shield_this_is_a_shield
```

This fixes the interpreter to stop giving that error, like the linter is already doing.


ysugimoto (approved):
Thanks, it is caused by https://github.com/ysugimoto/falco/pull/197 changes.

PR #213, opened 2023-11-14
Don't generate error for ratelimit functions

MasonM:
Currently, calling any of the `ratelimit.*` functions triggers the error `[ratelimit.*] Not implemented`. This means it's impossible to unit test any code that calls one of those functions. Other unimplemented functions (e.g. `resp.tarpit()`) instead return some default value, which is better behavior IMO, so I did the same thing here.


ysugimoto (approved):
hundred percent agree.

PR #216, opened 2023-11-15
Fix boolean operator precedence

MasonM:
The `&&` operator has higher precedence than `||`, so `true || false && false` should evaluate to `true`, but it was evaluating to `false`. Proof: https://fiddle.fastly.dev/fiddle/e184b801


ysugimoto (approved):
This fix is really important improvement, thanks!

PR #221, opened 2023-12-06
State change fixes

richardmarshall:
Fixes a few issues in the interpreter related to changing state.

1. When moving to the HIT state the context variables were not being properly set up for the new state.
2. Changing states from RECV to HASH wasn't logged
3. A non-existent return from vcl_hash was reported as an error. This is permitted by Fastly and is treated the same as explicitly returning `return(hash)`.
4. The `bereq` was not reset on a restart. This can cause unexpected behavior as changes to the `req` object after the restart aren't included in `bereq` when the backend request is made.


interpreter/interpreter.go
```diff
@@ -67,6 +69,7 @@ func (i *Interpreter) SetScope(scope context.Scope) {
 func (i *Interpreter) restart() error {
 	i.ctx.Restarts++
 	i.Debugger.Message(fmt.Sprintf("Restarted (%d) time", i.ctx.Restarts))
+	i.ctx.BackendRequest = nil
```
ysugimoto, line 72: Probably we need to reset request/response related pointers like:

- `i.ctx.Object`
- `i.ctx.BackendResponse`
- `i.ctx.Response`

ysugimoto (approved):
Thanks, and I put a bit comment.

interpreter/interpreter.go
ysugimoto: Good point.

I agree with you. Apparently, we don't need these nil checks 👍

richardmarshall: Done.

ysugimoto (approved):
Thanks, this should be merged.

PR #223, opened 2023-12-10
Propagate bare returns out of nested block statements

### LINKED ISSUE #222: [BUG] Bare returns do not propagate out of nested block statements
**Describe the problem**

When a subroutine has nested block statements, either direct block statements `{ // additional statements }` or other statements that include block statements such as if-statements, a bare return in the nested block does not propagate out and end execution of the subroutine. Evaluation of the subroutine in the nested block stops at the return statement and continues at the point after the nested block statement.

**VCL code that cause the problem**

```
backend localhost {
  .host = "localhost";
}

sub test_recv {
  if (true) {
    set req.http.step = "1";
    return; // subroutine evaluation should stop here
  }
  // this code should not be evaluated.
  set req.http.step = "2";
}

sub vcl_recv {
  call test_recv;
  if (req.http.step == "1") {
    log "test passed";
  } else {
    log "test failed: " req.http.step;
  }
  error 200 "OK";
}
```

Simple falco test case which shows the problem:
```
sub test_return_recv {
	assert.equal(req.http.step, "1");
}
```

Falco test output

```
Running tests... Done.
 FAIL  test_cases/return_statement/return.test.vcl
  ●  [RECV] test_return_recv

    Assertion Error: Assertion error: expect=1, actual=2
    Actual Value: 2

    2|     testing.call_subroutine("vcl_recv");
    3|     assert.equal(req.http.step, "1");
    4| }

0 passed, 1 failed, 1 total, 1 assertions
```

**Expected behavior**
A return statement in a nested block statement of a subroutine should stop execution of the subroutine.

**Desktop (please complete the following information):**
 - OS: linux
 - Shell: bash

**Additional context**

richardmarshall:
Adds the ability for the statement interpreter to differentiate between a block statement that was completed without a return and one that had a bare return `return;`.  This ensures that evaluation of a subroutine will end at the bare return and not continue after the nested block statement.

Additionally adds support for handling evaluation of bare block statements.

```
sub test {
  {
    log "I wasn't executed previously";
  }
}
```

Resolves #222

### DIFF
--- a/interpreter/interpreter_test.go
+++ b/interpreter/interpreter_test.go
@@ -60,7 +60,7 @@ func assertInterpreter(t *testing.T, vcl string, scope context.Scope, assertions
 			return
 		}
 		if diff := cmp.Diff(val, v); diff != "" {
-			t.Errorf("Value asserion error, diff: %s", diff)
+			t.Errorf("Value assertion error, diff: %s", diff)
 		}
 	}
 
@@ -77,7 +77,7 @@ func assertValue(t *testing.T, name string, expect, actual value.Value) {
 		return
 	}
 	if diff := cmp.Diff(expect, actual); diff != "" {
-		t.Errorf("Value asserion error, diff: %s", diff)
+		t.Errorf("Value assertion error, diff: %s", diff)
 	}
 }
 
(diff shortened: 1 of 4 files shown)


ysugimoto (approved):
Looks good.

PR #227, opened 2023-12-16
Update README with GitHub Actions support [ci skip]

ain:
Pushed the README update in case some are looking for easy ways to validate their PRs on GitHub Actions with Falco. 

We're using it in production, thanks for your great work!


ysugimoto (approved):
Thanks for your cooperation!

PR #228, opened 2023-12-19
User defined function fixes

richardmarshall:
## Unable to call user defined functions

The primary issue this PR addresses is the fact that currently user defined functions with return values can't be called in tests or in the simulator. The interpreter returns an error `Function <name> is not defined`.

### Example

`main.vcl`
```
backend dummy {
    .host = "localhost";
}

// @scope: recv
sub bool_fn BOOL {
    return true;
}

sub vcl_recv {
    set req.http.foo = "0";
    if (bool_fn()) {
        set req.http.foo = "1";
    }
}
```

`main.test.vcl`
```
// @scope: recv
sub test {
    assert.equal(req.http.foo, "1");
}
```

`falco test main.vcl`
```
Running tests... Done.
 FAIL  main.test.vcl
  ●  [RECV] test

    Function bool_fn is not defined

    1| sub test {
    2|     testing.call_subroutine("vcl_recv");
    3|     assert.equal(req.http.foo, "1");

0 passed, 1 failed, 1 total, 0 assertions
```

While fixing this issue I discovered some related issues with function handling.

### Boolean values stored with type name inconsistent with VCL BOOL

Returning a bool value from a function with a BOOL return type caused a runtime exception due to the mismatch between the return type name `BOOL` and the bool value type name `BOOLEAN`.

```
[RuntimeException] Invalid return type, expects=BOOL, but got=BOOLEAN in main.vcl at line: 7, position: 5
```

### Builtin functions with ID arguments

Builtin functions that take an ID as their first argument failed when statement called.

```
sub vcl_recv {
    header.set(req, "foo", "bar");
}
```

```
[RuntimeException] Function header.set of 0 argument must be an Ident in main.vcl at line: 11, position: 5
```

### User defined functions cannot be statement called

VCL does not permit user defined functions to have VOID return type and so they can't be statement called.

```
sub foo BOOL { return true; }
sub vcl_recv {
    foo(); // should be an error
}
```

https://fiddle.fastly.dev/fiddle/c0c4010b

### ID is not a value type

VCL does not allow the ID type to be used as a variable or function return type.

```
sub foo ID { return req; } // should be an error
sub vcl_recv {
    declare local var.bar ID; // should be an error
    var.bar = req;
}
```

https://fiddle.fastly.dev/fiddle/e07db67d

### DIFF
--- a/interpreter/expression.go
+++ b/interpreter/expression.go
@@ -11,6 +11,7 @@ import (
 	"github.com/ysugimoto/falco/interpreter/function"
 	"github.com/ysugimoto/falco/interpreter/operator"
 	"github.com/ysugimoto/falco/interpreter/value"
+	"github.com/ysugimoto/falco/types"
 )
 
 func (i *Interpreter) IdentValue(val string, withCondition bool) (value.Value, error) {
@@ -194,6 +195,29 @@ func (i *Interpreter) ProcessIfExpression(exp *ast.IfExpression) (value.Value, e
 }
 
 func (i *Interpreter) ProcessFunctionCallExpression(exp *ast.FunctionCallExpression, withCondition bool) (value.Value, error) {
+	if sub, ok := i.ctx.SubroutineFunctions[exp.Function.Value]; ok {
+		if len(exp.Arguments) > 0 {
+			return value.Null, exception.Runtime(
+				&exp.GetMeta().Token,
+				"Function subroutine %s could not accept any arguments",
+				exp.Function.Value,
+			)
+		}
+		if _, ok := types.ValueTypeMap[sub.ReturnType.Value]; !ok {
+			return value.Null, exception.Runtime(
+				&sub.GetMeta().Token,
+				"subroutine %s has invalid return type %s",
+				sub.Name,
+				sub.ReturnType,
+			)
+		}
+		// Functional subroutine may change status
+		v, _, err := i.ProcessFunctionSubroutine(sub, DebugPass)
+		if err != nil {
+			return v, errors.WithStack(err)
+		}
+		return v, nil
+	}
 	fn, err := function.Exists(i.ctx.Scope, exp.Function.Value)
 	if err != nil {
 		return value.Null, errors.WithStack(err)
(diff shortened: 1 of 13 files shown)


ast/vcl_types.go
```diff
@@ -14,6 +26,14 @@ func (i *Ident) GetMeta() *Meta { return i.Meta }
...
 	return i.LeadingInlineComment() + i.Value + i.TrailingComment()
 }
+func (i *Ident) IsValueType() bool {
+	for _, v := range valueTypes {
+		if v == i.Value {
+			return true
+		}
+	}
+	return false
+}
```
ysugimoto, line 36: Suggestion
Ident will express kind of types so this function should be a helper function in `interpreter` packages somewhere instead of `ast` method.

ast/vcl_types_test.go
```diff
@@ -0,0 +1,35 @@
...
+			expect: false,
+		},
+	}
+
+	for _, test := range tests {
+		if test.ident.IsValueType() != test.expect {
+			t.Errorf("%s: expect %t got %t", test.name, test.expect, !test.expect)
+		}
+	}
+}
```
ysugimoto, line 35: This tests should also move to `interpreter` package somewhere

interpreter/expression.go
```diff
@@ -194,6 +194,29 @@ func (i *Interpreter) ProcessIfExpression(exp *ast.IfExpression) (value.Value, e
...
+				sub.ReturnType,
+			)
+		}
+		// Functional subroutine may change status
+		v, _, err := i.ProcessFunctionSubroutine(sub, DebugPass)
+		if err != nil {
+			return v, errors.WithStack(err)
+		}
+		return v, nil
+	}
```
ysugimoto, line 219: Nice catch 👍

ysugimoto (commented):
Thanks for the lots of bug fixes!
I put some suggestions about a valueTypes.

ast/vcl_types.go
richardmarshall: Realized there's already a set of valid value types defined in `linter/helper.go`, should have remembered that since I removed `ID` from it. 😅 Makes sense to me to reuse that same map for both the linter and the interpreter. I've moved that map to the types package and used that instead of the IsValueType method. Thoughts?

ysugimoto (approved):
Looks good to me 👍 I'll merge this.

PR #231, opened 2023-12-23
Fix not_set value related implementation

ysugimoto:
This PR includes important fix.

## Change treating way of `NotSet` value in `STRING` and `IP` VCL type

VCL has a `NotSet` mechanism in `STRING` and `IP` types and it becomes the following situations:

- Local variable declaration
- Getting no existing HTTP header

See documentation in https://developer.fastly.com/reference/vcl/types/#not-set

For example:

```vcl
sub vcl_recv {
  declare local var.S STRING; // <= NotSet on declaration
  if (req.http.Foo) { // <= NotSet on getting no existing HTTP header
    ...
  }
}
```

The `NotSet` value is different from an empty value:

1. Comparison of `NotSet` explicitly `false`
2. Compare with empty value should be false

For example:

```vcl
sub vcl_recv {
  declare local var.S STRING; // NotSet
  if (!var.S) { // true
    set req.http.NotSet = "yes";
  }
  if (var.S == "") { // false
    set req.http.NotSetWithEmpty = "yes";
  }

  set var.S = ""; // Empty
  if (!var.S) { // false
    set req.http.Empty = "yes";
  }
  if (var.S == "") { // true
    set req.http.EmptyWithEmpty = "yes";
  }
}
```

See in fiddle: https://fiddle.fastly.dev/fiddle/da7d3fc3

And, in order to assert value is `NotSet`, implemented `assert.is_notset()` testing function.


interpreter/variable/local.go
```diff
@@ -66,16 +66,12 @@ func (v LocalVariables) Set(name, operator string, val value.Value) error {
...
-	if _, ok := v[name]; !ok {
-		return errors.WithStack(fmt.Errorf(
-			"Undeclared variable %s", name,
-		))
-	}
-	delete(v, name)
-	return nil
+	return errors.WithStack(fmt.Errorf(
+		"Cannot unset local variable %s", name,
+	))
```
ysugimoto, line 76: Fastly recognized that has wrong documentation.
https://developer.fastly.com/reference/vcl/variables/#user-defined-variables

We cannot unset the local variable.

interpreter/statement.go
```diff
@@ -430,7 +430,7 @@ func (i *Interpreter) ProcessIfStatement(stmt *ast.IfStatement, ds DebugState) (
 			return state, nil
 		}
 	case *value.String:
-		if t.Value != "" {
+		if !t.IsNotSet {
```
MasonM, line 433: Shouldn't the "else if" section have this change too? https://github.com/ysugimoto/falco/blob/105b9118206a212abc75144af629839130f8839d/interpreter/statement.go#L469-L471

ysugimoto: Good point. I forget to fix here, thanks

ysugimoto: Fixed it, and also fixed in if expression.

MasonM (approved):

PR #234, opened 2023-12-25
tmp: suppress notset output

ysugimoto:
This is a temporal fix, on the concatenation string, NotSet variable should be treated as the empty string.

However, in Fastly, sometimes NotSet string concatenation is treated as a "(null)" string.
We can see the behavior at https://fiddle.fastly.dev/fiddle/e9852716

Temporary it's okay on this behavior but to adjust Fastly properly, we may need to more fixation for this problem.

### DIFF
--- a/interpreter/value/value.go
+++ b/interpreter/value/value.go
@@ -71,7 +71,7 @@ type String struct {
 }
 
 func (v *String) String() string {
-	// Temporaly comment out to supress not set output
+	// Temporaly comment out to suppress not set output
 	// if v.IsNotSet {
 	// 	return "(null)"
 	// }
--- a/interpreter/value/value.go
+++ b/interpreter/value/value.go
@@ -71,9 +71,10 @@ type String struct {
 }
 
 func (v *String) String() string {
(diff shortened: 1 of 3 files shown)


interpreter/value/value.go
```diff
@@ -71,9 +71,10 @@ type String struct {
 }
 
 func (v *String) String() string {
-	if v.IsNotSet {
-		return "[not set]"
-	}
+	// Temporaly comment out to suppress not set output
```
richardmarshall, line 74: Nit, typo: Should be `Temporarily`, same on line 97.

richardmarshall (approved):
Seems like a reasonable first step to me, moving the handling of the "(null)" case to the operators / functions which treat unset strings that way fits with how the documentation describes the behavior. Unfortunately overall the Fastly documentation about this is rather vague.

> Various functions and operators treat unset strings differently; some render them as the empty string, and some as "(null)". This handling is a property of the function (or operator), rather than a property of the STRING type.

Would be nice if they gave a list of where unset strings are handled as "" vs "(null)".

PR #238, opened 2023-12-30
Add support for switch statements

richardmarshall:
While not documented on the Fastly developer portal the Fastly VCL variant supports switch statements.

Simple example:
```vcl
switch (req.http.foo) {
case "bar":
  log "matched bar";
  break;
case "baz":
  log "matched baz";
  fallthrough;
default:
  log "default";
  break;
}
```

Experimentation shows the following grammar ([WSN](https://en.wikipedia.org/wiki/Wirth_syntax_notation)) for the switch statement:
```flex
SwitchStmt = "switch" "(" Control ")" "{" CaseClause { CaseClause } "}" .

// Control value cast to STRING before comparison with case tests.
// User defined function calls must return STRING.
// Built-in function calls must return a value type.
Control = valueTypeIdent | functionCall | stringLiteral | boolLiteral .

CaseClause = SwitchCase ":" StatementList ( "break" | FallThroughStmt ) ";" .

// Case tests must be unique
SwitchCase = "case" Test | Default .

// There can be at most one default case and it can appear anywhere in the case list.
Default = "default" .

StatementList = { statement ";" } .

// Continue execution in the next case clause. May not be used in the final case.
FallThroughStmt = "fallthrough" .

// Tests are either string literals or regular expressions.
Test = stringLiteral | RegexMatch .
RegexMatch = "~" stringLiteral . 

valueTypeIdent = /* builtin or user defined value type variable */ . 
functionCall = /* builtin or user defined function call expression */ .
stringLiteral = /* VCL string literal */ .
boolLiteral = /* VCL bool literal */ .
statement = /* Any VCL statement that is valid in a block statement body. */ .
```

Fiddle showing various aspects of how switch statements behave: https://fiddle.fastly.dev/fiddle/0226bcbb

Since switch statements are valid Fastly VCL, Falco should handle them correctly. This PR adds support for parsing, interpreting, and linting switches.

The linting rules are very basic to keep the scope of the PR reasonable.


ast/switch_statement.go
```diff
@@ -0,0 +1,30 @@
+package ast
+
+import "bytes"
+
+type SwitchStatement struct {
+	*Meta
+	Control Expression
+	Cases   []*CaseStatement
+	Default int
```
ysugimoto, line 9: `Default` seems that `switch` statement has `default` or not and indicates index of `Cases` field as default statement but I would suggest it should be `*CaseStatement` and set on parse `DEFAULT` token particularly.

ast/case_statement.go
```diff
@@ -0,0 +1,53 @@
...
+func (s *CaseStatement) IsFallthrough() bool {
+	_, ok := s.FinalStatement().(*FallthroughStatement)
+	return ok
+}
+func (s *CaseStatement) TestEqual(other *CaseStatement) bool {
+	if s.Test == nil || other.Test == nil || s.Test.Operator != other.Test.Operator {
+		return false
+	}
+	return s.Test.Right.String() == other.Test.Right.String()
+}
```
ysugimoto, line 31: `AST` must have only syntax tree and token representation so we should not have any logic here.
An `interpreter` package should implement this logic

interpreter/statement.go
```diff
@@ -115,6 +124,9 @@ func (i *Interpreter) ProcessBlockStatement(statements []ast.Statement, ds Debug
 			state := i.ProcessReturnStatement(t)
 			return state, DebugPass, nil
 
+		case *ast.FallthroughStatement:
+			return FALLTHROUGH, DebugPass, nil
+
```
ysugimoto, line 129: I'm wondering why we handle fallthrough statement here  - it should be done in `ProcessSwitchStatement` because the `fallthrough` statement is just an internal state in `swtich` statement, not an VCL statement.

ysugimoto (changes requested):
Very interesting.

lexer implementation seems to be okay but before seeing linter/interpreter implementation, we need to discuss about AST definition and switch statement specification.

For AST definition, I put some suggestions that may modify parser implemnentations. check them out, please.
For the switch statement specification, at least I confirmed at Fastly Fiddle:

## Could not write early `break` in nested statement

In `case` statement, the another `break` statement could not be present.

```vcl
declare local var.Control STRING;
set var.Control = "1";

switch (var.Control) {
  case "1":
    if (req.http.Control) {
      break;  // <- ERROR
    }
    set req.http.Control = var.Control;
    break;
} 
```
It seems to cause syntax error in fiddle: https://fiddle.fastly.dev/fiddle/fe2ac2c2

Probably the switch statement in VCL have another limitation - it's not like some programming language - we need to check any other cases.

ast/case_statement.go
```diff
@@ -0,0 +1,53 @@
...
+	return s.Statements[len(s.Statements)-1]
+}
+func (s *CaseStatement) IsBreak() bool {
+	_, ok := s.FinalStatement().(*BreakStatement)
+	return ok
+}
+func (s *CaseStatement) IsFallthrough() bool {
+	_, ok := s.FinalStatement().(*FallthroughStatement)
+	return ok
+}
```
ysugimoto, line 25: `AST` must have only syntax tree and token representation so we should not have any logic here.
An `interpreter` package should implement this logic

ast/switch_statement.go
richardmarshall: I implemented it this way with the index due to the fact that the default case can be fallen into and out of.

This is valid and will have log output from all 3 cases:
```
set req.http.foo = "1";
switch (req.http.foo) {
  case "1":
    log "case 1";
    fallthrough;
  default:
    log "default";
    fallthrough;
  case "2":
    log "case 2";
    break;
}
```

The default case could be stored outside of the case list, but it would complicate implementation of this fallthrough behavior.

richardmarshall:
> ## Could not write early `break` in nested statement
> In `case` statement, the another `break` statement could not be present.

Ah thank you, missed that rule when validating the implementation.

ast/case_statement.go
richardmarshall: Ok makes sense. I was looking at the AST type as just a convenient place to put some helper methods that would be used by the parser when validating the cases. Is your view that these kinds of checks should be done in the interpreter (and by extension in the linter) rather than in the parser?

Specifically checking for duplicate cases and ensuring the final case ends with a `break`.

interpreter/statement.go
richardmarshall: It just seemed to me to be a straightforward way to indicate that a case should fall through since it doesn't require storing that information into the case AST struct or needing to check what the final statement was in the cases statement list. But I'll rework this to not have any internal switch logic in `ProcessBlockStatement`.

ast/switch_statement.go
ysugimoto: Understood, to think of `fallthrough` behavior, your way would be better.

ast/case_statement.go
richardmarshall: Removed the helper methods from the CaseStatement type. To start I left the syntax validation in the parser if you'd prefer for that validation to be moved to the interpreter I can do that as well.

interpreter/statement.go
richardmarshall: Removed the handling of `fallthrough` from `ProcessBlockStatement`.

ysugimoto (changes requested):
Looks good entirely, and put final comments to be approved, thanks!

interpreter/statement.go
```diff
@@ -494,3 +503,94 @@ func (i *Interpreter) ProcessIfStatement(stmt *ast.IfStatement, ds DebugState) (
...
+			return NONE, errors.WithStack(err)
+		}
+		return state, nil
+	}
+
+	return NONE, nil
+}
+
+func (i *Interpreter) ProcessCaseStatement(stmt *ast.SwitchStatement, offset int,
+	control value.Value, isFallthrough bool, ds DebugState) (State, bool, error) {
```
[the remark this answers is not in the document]
richardmarshall: Done

interpreter/statement.go
```diff
@@ -494,3 +503,94 @@ func (i *Interpreter) ProcessIfStatement(stmt *ast.IfStatement, ds DebugState) (
...
+
+	var matched bool
+	if stmt.Default == offset || isFallthrough {
```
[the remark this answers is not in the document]
richardmarshall: Done.

interpreter/statement.go
```diff
@@ -494,3 +503,94 @@ func (i *Interpreter) ProcessIfStatement(stmt *ast.IfStatement, ds DebugState) (
...
+		if fn, ok := i.ctx.SubroutineFunctions[fnCall.Function.Value]; ok {
+			if fn.ReturnType.Value != string(value.StringType) {
+				return NONE, errors.WithStack(exception.Runtime(
+					&fnCall.Token,
+					"user defined function has invalid return type (%s) for switch, must be STRING",
+					fn.ReturnType.Value,
+				))
+			}
+		}
+	}
```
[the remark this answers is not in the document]
richardmarshall: We don't want to return early in this section if the function isn't found as this block is for validating the return type only if the function is user defined. If it doesn't exist in `SubroutineFunctions` it could just mean the function is a builtin. The call to `i.ProcessExpression` after this section will handle the case of the function not being defined.

That said this could still have one layer of nesting removed.
```
fn, ok := i.ctx.SubroutineFunctions[fnCall.Function.Value]
if ok && fn.ReturnType.Value != string(value.StringType) {
   // ...
}
```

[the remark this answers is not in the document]
ysugimoto: Do you mean the `ast.FunctionCallStatement` is a possibility to be a built-in function? if so, you are right

PR #242, opened 2024-01-22
support return value in block statement and if statement

### LINKED ISSUE #241: [BUG] cannot return INTEGER from within an if branch in a subroutine
Please fill all **required** sections

### Kind of proposals

- []Parser
- []Linter
- [x]Simulator
- []Testing

### Describe the problem

It seems to be kind of a specific bug, but I cannot have a "return" statement within an "if" branch of a subroutine that returns an INTEGER. This is hard to understand with words, so please have a look at the following simple code that better illustrates what I mean.

```
$ cat main.vcl

sub compute INTEGER {
  // pb seems to be around here
  if (true) {
    return 2;
  }
  return 3;
}

sub vcl_recv {
  declare local var.myint INTEGER;
  set var.myint = compute();
}
```

And when I launch the simulator and refresh my browser at http://localhost:3124, I have this error:

```
$ falco simulate main.recv
...
Failed to assign value to var.myint, Invalid assignment for INTEGER type, got NULL
```

Here is a version working in Fastly Fiddle.

https://fiddle.fastly.dev/fiddle/d4bd0467

Also, if I remove the `if` in `compute`, it works:

```
sub compute INTEGER {
  return 2;
}

sub vcl_recv {
  declare local var.myint INTEGER;
  set var.myint = compute();
}
```

### Expected behavior

The above should not raise an error.

 - OS: ubuntu 22
 - Shell: bash

ysugimoto:
This PR fixes #241 

On functional subroutine, it is a possibility to return value inside nested block statement or if statement.
To support this behavior, I changed that a BlockStatement returns either `value.Value` or `State` - it depends on the interpreter context.

The interpreter provides its context to block statements that are inside the functional subroutine or not, and switches `ReturnStatement` behavior that returns `value.Value` or `State`.


MasonM (approved):

richardmarshall:
Didn't see this before it was merged but the switch statement handling got missed.

This code for example still shows the previous behavior in the interpreter. The return value in this case is "0" when it should be "1".
```vcl
sub test STRING {
   set req.http.foo = "1";
   switch (req.http.foo) {
   case "1":
     return "1";
     break;
   }
   return "0";
}
```

Also a bare nested block seems to not be handled correctly:
```
sub test STRING {
   {
     return "1";
   }
}
```

When trying to call that function the interpreter complains with the following error:
```
[RuntimeException] Functioncal subroutine test did not return any values in main.vcl at line: 1, position: 1
```

This does work on Fastly: https://fiddle.fastly.dev/fiddle/871498a6

I'll make a PR to fix this later today.

PR #244, opened 2024-01-23
Value returns from switch and bare block statements

richardmarshall:
Fix missing support for value returns from switch statements and bare block statements.

Related to #242 and #241


ysugimoto (approved):
Looks good! Thanks 👍

PR #247, opened 2024-01-24
Add assert.not_subroutine_called

bungoume:
This PR will add `assert.not_subroutine_called` asserter.

* assert.not_subroutine_called Assert subroutine has not called in testing subroutine.

I have a test case where I want to verify that a function has never been called.


ysugimoto (approved):
Looks good!

Issue #237, opened 2023-12-26
[BUG] (simulator) Behavior when multiple lines have the same header.


bungoume:
**Describe the problem**
Fastly seems to exhibit a special behavior when there are multiple lines with the same header.
https://fiddle.fastly.dev/fiddle/6221d4db

While the specific behavior needs to be verified accurately, I'll write test cases based on the current understanding.

* Only the Header listed at the beginning is utilized:
    * When assigned, the value from the first instance is used.
    * Conditions such as 'if' statements use the value from the initial entry.
    * Writing using : such as set req.http.HEADER:V = ""; overwrites the first header entry, erasing subsequent entries from the second onwards.
* Header sequence reverses during the fetch process.
* Regardless of the presence of multiple lines, using unset removes all lines.
* Even with multiple lines, using set retains only that single line.
* At the stage of the Response from Fastly, multiple lines consolidate into a single header separated by commas.

**VCL code that cause the problem / reproduceable**
https://fiddle.fastly.dev/fiddle/aa00bc57

```vcl
// @scope: recv
// @suite: ADD header(add-add-add) BUGGY
sub test_recv {
    add req.http.VALUE = "V1";
    add req.http.VALUE = "V2";
    add req.http.VALUE = "V3";
    # assert.equal(req.http.VALUE, "V1, V2, V3");  # request upstream with 3 line headers
    # assert.equal(resp.http.VALUE, "V1, V2, V3");  # response contains single line header with three values.

    set req.http.MESSAGE = req.http.VALUE;  # set first header value
    assert.equal(req.http.MESSAGE, "V1");
}

// @scope: recv
// @suite: ADD header(set-add-add) BUGGY
sub test_recv {
    set req.http.VALUE = "V1";
    add req.http.VALUE = "V2";
    add req.http.VALUE = "V3";
    # assert.equal(req.http.VALUE, "V1, V2, V3");  # request upstream with 3 headers

    set req.http.MESSAGE = req.http.VALUE;  # set first header value
    assert.equal(req.http.MESSAGE, "V1");
}

// @scope: recv
// @suite: ADD header(add-add-set)
sub test_recv {
    add req.http.VALUE = "V1";
    add req.http.VALUE = "V2";
    set req.http.VALUE = "V3";
    assert.equal(req.http.VALUE, "V3");  # 1 header

    set req.http.MESSAGE = req.http.VALUE;
    assert.equal(req.http.MESSAGE, "V3");
}

// @scope: recv
[...]

Issue #236, opened 2023-12-26
[BUG] (simulator) Editing header values using colons


bungoume:
**Describe the problem**
It seems that the result of rewriting the header value using colons differs from Fiddler's output.
Could you please review the following case?

**VCL code that cause the problem / reproduceable**
https://fiddle.fastly.dev/fiddle/da790883

```vcl
// @scope: recv
// @suite: SET VARS VALUE
sub test_recv {
    set req.http.VARS = "";
    set req.http.VARS:VALUE = "V";
    assert.equal(req.http.VARS, "VALUE=V");
}

// @scope: recv
// @suite: SET NOT-INITIALIZED VARS VALUE
sub test_recv {
    set req.http.VARS:VALUE = "V";
    assert.equal(req.http.VARS, "VALUE=V");
}

// @scope: recv
// @suite: SET MULTIPLE VARS VALUE
sub test_recv {
    set req.http.VARS = "";
    set req.http.VARS:VALUE = "V";
    set req.http.VARS:VALUE2 = "V2";
    assert.equal(req.http.VARS, "VALUE=V, VALUE2=V2");
}

// @scope: recv
// @suite: SET EMPTY VARS VALUE
sub test_recv {
    set req.http.VARS = "";
    set req.http.VARS:VALUE = "";
    assert.equal(req.http.VARS, "VALUE");
}

// @scope: recv
// @suite: SET MULTIPLE EMPTY VARS VALUE
sub test_recv {
    set req.http.VARS = "";
    set req.http.VARS:VALUE = "";
    set req.http.VARS:VALUE2 = "";
    assert.equal(req.http.VARS, "VALUE, VALUE2");
}

// @scope: recv
// @suite: UNSET VARS ALL VALUE
sub test_recv {
    set req.http.VARS = "";
    set req.http.VARS:VALUE = "V";
    unset req.http.VARS:VALUE;
    assert.is_notset(req.http.VARS);
}

// @scope: recv
// @suite: UNSET VARS VALUE
sub test_recv {
    set req.http.VARS = "";
    set req.http.VARS:VALUE = "V";
    set req.http.VARS:VALUE2 = "V2";
    unset req.http.VARS:VALUE;
    assert.equal(req.http.VARS, "VALUE2=V2");
}

// @scope: recv
// @suite: OVERRIDE VARS VALUE
sub test_recv {
    set req.http.VARS = "";
    set req.http.VARS:VALUE = "V";
    set req.http.VARS:VALUE = "O";
[...]

PR #248, opened 2024-01-27
Support NotSet string behavior in falco

### LINKED ISSUE #235: [BUG] (testing) set req.http.VALUE = "" should not be [not set]
**Describe the problem**
`set req.http.VALUE = "";` and`unset req.http.VALUE;` should be different, but they seem to behave the same in falco v1.3.0.

**VCL code that cause the problem / reproduceable**
see https://fiddle.fastly.dev/fiddle/2e7f4a1c

```vcl
// @scope: recv
// @suite: UNSET header
sub test_recv {
    set req.http.VALUE = "V";
    unset req.http.VALUE;
    assert.is_notset(req.http.VALUE);

    set req.http.MESSAGE = req.http.VALUE;
    assert.equal(req.http.MESSAGE, "(null)");
}

// @scope: recv
// @suite: EMPTY header
sub test_recv {
    set req.http.VALUE = "V";
    set req.http.VALUE = "";
    assert.equal(req.http.VALUE, "");

    set req.http.MESSAGE = req.http.VALUE;
    assert.equal(req.http.MESSAGE, "");
}

// @scope: recv
// @suite: NOTSET header
sub test_recv {
    set req.http.VALUE = "V";
    set req.http.VALUE = req.http.NOTSET;
    assert.is_notset(req.http.VALUE);

    set req.http.MESSAGE = req.http.VALUE;
    assert.equal(req.http.MESSAGE, "(null)");
}
```

**Additional context**
https://github.com/bungoume/falco-vcl-empty-test/blob/main/tests/header.test.vcl

ysugimoto:
Fixes #235 #236 #237

I apologize this PR includes massive changes.

This PR implements `NotSet` string support which is described [here](https://developer.fastly.com/reference/vcl/types/string/). To support this behavior, I had to make fundamental changes to HTTP header manipulation:

## Implement original HTTP-related struct

To simulate Fastly HTTP-related behavior conveniently, I implemented the following structs originally:
  - `http.Request`
  - `http.Response`
  - `http.Header`
  - cookie retrieving function
 
These are equivalent to Golang's `net/http` package and can convert each other through the utility function.
You can see the above structs as `flchttp` aliased import.

Particularly, `http.Header` implements Fastly header manipulation feature, colon-separated syntax, multiple header behavior, and adding `(null)` string to the actual HTTP request when NotSet string is assigned.

## Implement more string value type

I implemented `LenientString`, which keeps assigned/concatenated values inside and can support to output `(null)` string if the value is entirely NotSet.

According to declare the new value type, I needed to change the entire expression evaluation stuff, assign, operator, built-in function, etc.

After this implementation, we need to consider whether the string type is either `*value.String` or `*value.LenientiString`, then we can use `value.GetString(value.Value)` as wrap them transparently.

## Implement new transport package

This is not related to NotSet-related things but the actual origin requests feature handles this package.
This package includes creating Golang's `http.Request` pointer from `flchttp.Request`, making director, and so on.
Please look at the implementation.

This change should bump the major version I think.

Again, sorry for the huge changes, please take a look at the changesets with having the above perspective and perhaps you have unclear points, then feel free to put comments to diff, and let's discuss it, thanks.


richardmarshall:
Definitely a lot to look through, will start reviewing the changes in the next day or two when I have a bigger block of free time.

examples/testing/notset_string/notset_string.test.vcl
richardmarshall, line 1: Started looking through the PR this morning and the test assertions in this file stood out to me.

For example in this test:
```
// @scope: recv
// @suite: UNSET header(add-add-unset)
sub test_recv {
    set req.http.VALUE = req.http.NOTSET;
    assert.is_notset(req.http.VALUE);

    set req.http.MESSAGE = req.http.VALUE;
    assert.equal(req.http.MESSAGE, "(null)");
}
```

What seemed odd to me was the fact that `assert.equal(req.http.MESSAGE, "(null)");` was passing when the header is unset. A condition like `req.http.MESSAGE == "(null)"` in an if statement for example should evaluate to `false`. So `assert.equal` should seemingly also reject this case.

Put together a fiddle that reproduces the sequence of `set`s from the test and inspecting the results.
https://fiddle.fastly.dev/fiddle/d1c67281

The log output of that fiddle is
```
req.http.VALUE
not set
logged value: (null)
req.http.MESSAGE
not set
logged value: (null)
```

Both the VALUE and MESSAGE headers are "not set" and don't equal "(null)" in the context of an if statement condition. But what I found interesting is that the logged value is "(null)". That fact started a whole other path of looking into the behavior of all this and I ran out of time to be able to get back to reviewing the PR.

I started putting together another fiddle with as many permutations of usage of unset header and local string variables I could think of to use as a basis for comparing the falco not set string behavior with what Fastly is doing. Will share that when I have some time to finish it and will get back to reading through more of the PR soon.

ysugimoto: Yes, you are right - this looks strange to us but it simulates Fastly's behavior.
The `NotSet` string evaluation depends on the context, sometimes empty string, and sometimes "(null)", and unfortunately Fastly VCL developer also could not grab all cases that the `NotSet` string could evaluate as "(null)".

However, I'd understand the cases following:

1. on `log`, `synthetic`,  and `synthetic_base64` statements, message string includes NotSet will evaluate as "(null)"
2. on making actual HTTP headers that send to the origin,  the header value string includes NotSet will evaluate as "(null)"
3. on assigning into `STRING` local variable like `var.SomeString` will evaluate as `empty string`
4. In other cases, as you pointed out if expression, `NotSet` string will be evaluated as `falsy` and any comparison operation should be treated like NULL.
5. built-in function - some functions will be treated as "(null)" but nobody is figuring it out now.

Probably there are more cases where the `NotSet` string is evaluated as "(null)" hence this PR includes the features that I only understand.

richardmarshall: Didn't see this when writing up the comment with the test case fiddle/tests yesterday. I had missed the `synthetic` statements when thinking through various places where not set values would need to be considered.

You're right `synthetic` does behave like the `log` statement and always expands not set values to `(null)`.

I'm not seeing the `(null)` behavior with `synthetic.base64` though. The response body in a fiddle with `synthetic.base64 req.http.unset` seems to be empty. Likewise doing a string concat expression with that statement `synthetic.base64 req.http.unset + "aGkK"` results in the body `hi` without the `(null)` expansion for the unset header.

richardmarshall:
The `LenientString`/`VString` types are handling two use-cases; one for handling the `(null)` expansion, and the other to store a collection of values for the header object representation (`req.http.foo:bar`).

## Multiple values

Since the only use case for the multiple values is the `Header` value KV object feature it feels like a detail that should be internal to the `Header` type rather than part of the interpreter type system.

## `(null)` expansion

Again this feels like functionality that should be internal to the statement  (`set`, `add`, `log`, `synthetic`) / expression (concatenation) handling rather than part of the STRING/IP value types or a wrapper for them in this case with the `LenientString` type. Since there are contextual details that need to be taken into account for when to perform expansion and when not to there will inevitably need to be logic in the expression parsing stage so it seems to me to make more sense for this logic to live there rather than in a value type.

Took a very quick stab at handling just the `(null)` expansion without making changes to the `String` or `IP` value type handling.

https://github.com/richardmarshall/falco/commit/3212a486de66f7426a0e516b2ae39da1ed17dae3

This approach is able to pass all the [tests](https://github.com/richardmarshall/falco/commit/3212a486de66f7426a0e516b2ae39da1ed17dae3#diff-594bc1daf7fdb16add8499643bccc697cf1d3f403520a95e333fe2583f075ad4) and limits the scope of changes needed to support the `(null)` behavior considerably compared to using the `LenientString` / `VString` model.

```
rgm@penguin:~/src/github.com/richardmarshall/falco$ make linux_amd64 && ./dist/falco test examples/testing/not_set/main.vcl 
GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build \
                 -ldflags "-X main.version=dev" \
cd ./dist/ && cp ./falco-linux-amd64 ./falco && tar cfz falco-linux-amd64.tar.gz ./falco
Running tests... Done.
 PASS  /home/rgm/src/github.com/richardmarshall/falco/examples/testing/not_set/not_set.test.vcl
  ✓ [RECV] Set STRING var the result of concatenation of an unset STRING var and unset STRING var
  ✓ [RECV] Set STRING var the result of concatenation of a string literal and unset STRING var
  ✓ [RECV] Set STRING var the result of concatenation of a string literal and unset IP var
  ✓ [RECV] Set STRING var the result of space concatenation of an unset STRING var and unset STRING var
  ✓ [RECV] Set STRING var the result of space concatenation of a string literal and unset STRING var
  ✓ [RECV] Set STRING var the result of space concatenation of a string literal and unset IP var
  ✓ [RECV] Set STRING var the result of concatenation of a string literal and unset header
  ✓ [RECV] Set STRING var the result of space concatenation of a string literal and unset header
  ✓ [RECV] Set header the result of concatenation of a string literal and unset STRING var
  ✓ [RECV] Set header the result of concatenation of a string literal and unset IP var
  ✓ [RECV] Set header the result of concatenation of a string literal and header
  ✓ [RECV] Add header the result of concatenation of a string literal and unset header
  ✓ [RECV] Add header the result of concatenation of a string literal and unset STRING var
  ✓ [RECV] Add header the result of concatenation of a string literal and unset IP var
  ✓ [RECV] Set header an unset STRING var
  ✓ [RECV] Set header an unset IP var
  ✓ [RECV] Set header an unset header
  ✓ [RECV] Add header an unset header
  ✓ [RECV] Add header an unset STRING
  ✓ [RECV] Add header an unset IP
  ✓ [RECV] Set STRING var an unset STRING var
  ✓ [RECV] Set STRING var an unset IP var
  ✓ [RECV] Set STRING var an unset header
  ✓ [RECV] Set STRING var result of json.escape with unset STRING argument
  ✓ [RECV] Set STRING var result of regsub with unset STRING replacement argument
  ✓ [RECV] Set STRING var result of json.escape with string concatenation argument containing an unset STRING
  ✓ [RECV] Use unset STRING variable as switch statement control
  ✓ [RECV] Use unset IP variable as switch statement control
  ✓ [RECV] Use unset header as switch statement control
  ✓ [RECV] Unset STRING variable in if statement condition
  ✓ [RECV] Unset header in if statement condition
  ✓ [RECV] Log statement with unset STRING variable
  ✓ [RECV] Log statement with unset IP variable
  ✓ [RECV] Log statement with unset header
34 passed, 0 failed, 34 total, 43 assertions
```

I'll start looking through the new transport package part of this PR soon.

interpreter/transport/backend.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+	if v, ok := prop.(T); ok {
+		return v, nil
+	}
+	return nil, errors.New("Type assertion failed")
+}
+
+func backendScheme(backend *value.Backend, override *config.OverrideBackend) (string, error) {
+	// scheme may be overrided by config
+	if override != nil {
```
richardmarshall, line 47: Since this pair of if statements are logically one thing it could be simplified into a single condition. `override != nil && override.SSL`.  There are a couple similar pairings in this file as well, lines 56/57 and 131/132.

Largely a style preference nit though.

richardmarshall (commented):
The transport package looks good, just a couple small comments.

interpreter/http/director.go
```diff
@@ -0,0 +1,247 @@
...
+			if !v.Backend.Healthy.Load() {
+				continue
+			}
+			for i := 0; i < v.Weight; i++ {
+				lottery[current] = index
+				current++
+			}
+		}
+
+		rand.New(rand.NewSource(time.Now().Unix()))
```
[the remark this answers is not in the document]
ysugimoto: Should be stored to variable and call `r.Intn` to be sure, thanks.

interpreter/transport/backend.go
```diff
@@ -0,0 +1,139 @@
...
+			break
+		}
+	}
+	if prop == nil {
+		return nil, nil
+	}
+	if v, ok := prop.(T); ok {
+		return v, nil
+	}
+	return nil, errors.New("Type assertion failed")
```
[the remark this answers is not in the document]
ysugimoto: `T` is just generics type info so now I improved message that user needs to set correct type for the property name of `key`.

interpreter/transport/backend.go
```diff
@@ -0,0 +1,139 @@
...
+			return HTTPS_SCHEME, nil
+		}
+	}
+	return HTTP_SCHEME, nil
+}
+
+func backendHost(backend *value.Backend, override *config.OverrideBackend) (string, error) {
+	// host may be overrided by config
+	if override != nil {
+		return override.Host, nil
```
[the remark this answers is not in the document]
ysugimoto: Yes, but I think it should be validated on config initialization. I'll do it on another issue.

interpreter/transport/backend.go
ysugimoto: Yes, but we could make it clear the `override.SSL` would be compared with a `bool` type.

ysugimoto:
@richardmarshall  https://github.com/richardmarshall/falco/commit/3212a486de66f7426a0e516b2ae39da1ed17dae3 is a great approach to consider the (null) string! could you send me a PR for this branch of `feat/re-implementation-for-header-value` please?

interpreter/http/director.go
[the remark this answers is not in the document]
ysugimoto: Ah you're right, we should remove this line to correct choosing backend by weight.

richardmarshall:
> @richardmarshall [richardmarshall@3212a48](https://github.com/richardmarshall/falco/commit/3212a486de66f7426a0e516b2ae39da1ed17dae3) is a great approach to consider the (null) string! could you send me a PR for this branch of `feat/re-implementation-for-header-value` please?

Sure will do, but probably won't have time for a few days.

richardmarshall:
I got started on rebasing my experimental changes onto this branch when I got derailed by discovering what seems to be a bug in Fastly's not_set handling for header fields. Realized I hadn't tested how header fields behave when assigned not_set values or are included in concatenation, etc. Through that I found that the length of the field name has an impact on the behavior.

```vcl
set req.http.test1 = req.http.unset;    // test1 header is a not_set string
set req.http.test2:a = req.http.unset;  // field a of the test2 header is a not_set string
set req.http.test3:ab = req.http.unset; // field ab of the test3 header is an empty string
```

Header fields with a key length of 1 behave differently than keys with length > 1 when the header is unset. Which doesn't make any logical sense so seems like a bug.

Fiddle showing this odd behavior:
https://fiddle.fastly.dev/fiddle/5dd31959

Anyway will get back to the rebase work soon.

richardmarshall:
Ended up running into another issue that opened up a whole new problem area for our compatibility with Fastly's header field implementation. Currently in our implementation setting the main header value to an encoded set of fields like `set req.http.foo = "a=1,b=2";` does not handle accessing the keys correctly. In this example accessing `foo:a` would return `1,b=2` and trying to access `foo:b` would return a not_set value.

Fastly parses the full set of fields and makes them accessible. I initially though supporting this seemed like it would be doable with some tweaks to the implementation in this PR until I found another weird case.

```vcl
set req.http.foo = "a=1,b=2";
log req.http.foo:a; // `1` as expected
log req.http.foo:b; // `2` as expected
set req.http.foo:c = ",d=3";
log req.http.foo:c; // `,d=3` as expected
log req.http.foo:d; // `3"` unexpected
```

Setting a field value that contains a field separator `,` and a new key/value pair then allows you to access the new key. This seems to indicate that Fastly isn't storing the header fields in a map of sorts and is instead parsing the main header value to extract the fields each time a field is accessed.

At this point I started putting together a set of validation tests to figure out as much about how fastly is dealing with these header fields. Found that the RFC-8941 specification for structured header values contained a number of test cases and used that as a starting point.

RFC - https://www.rfc-editor.org/rfc/rfc8941.html
Dictionary test cases - https://github.com/httpwg/structured-field-tests/blob/main/dictionary.json
Validation tests: https://github.com/richardmarshall/falco-validation-tests/blob/main/header_fields/main.test.vcl

I don't have a fiddle prepared confirming the Fastly results of all the tests but I can put that together soon.

Results for main and this PRs branch:

main:
```
5 passed, 36 failed, 41 total, 44 assertions
```

This PR:
```
2 passed, 39 failed, 41 total, 41 assertions
```

Took a quick stab at trying to handle field access in a way that didn't rely on storing the fields independently from the main header value. Started with an approach based on simple string splits on the field separator and was able to significantly improve the compatibility test results.

```
17 passed, 24 failed, 41 total, 70 assertions
```

But ran into challenges with properly dealing with field values that contain a `,`. Started over with a regex solution and was able to construct an expression that seems to properly match the behavior of Fastly and gets all the compatibility tests to pass.

```
41 passed, 0 failed, 41 total, 107 assertions
```

The regex pattern is hard to interpret and could likely be optimized for clarity but it handles the idiosyncrasies of the Fastly behavior.

https://github.com/ysugimoto/falco/compare/main...richardmarshall:falco:header_field_experiment

Additionally I did open a PR against this branch for my not_set handling #265.

ysugimoto:
Oh really, I also did not know such behavior in dealing with comma-separated header values.
So then it makes us confused about which `set` or `add` statements should we use... but your investigation would be correct, thanks

richardmarshall:
> Oh really, I also did not know such behavior in dealing with comma-separated header values. 

Yeah it was quite a surprise when I first saw it. Further experimentation found a number of additional unexpected or currently unaccounted for behaviors. I believe all of the ones I found are handled properly in the proof of concept implementation in #267 

> So then it makes us confused about which `set` or `add` statements should we use... 

Don't follow, what would be confusing?

ysugimoto:
This PR should be closed due to implement on #447

PR #254, opened 2024-02-06
Save/restore current subroutine locals when processing call statement

### LINKED ISSUE #253: [BUG] local variables not saved on subroutine call
### Kind of proposals

- [] Parser
- [] Linter
- [x] Simulator
- [x] Testing

### Describe the problem

When processing a subroutine the local variables of the outer scope are not saved/cleared on entry to the called subroutine. This allowed the nested subroutine to access the variables from the outer scope. Additionally when the called subroutine exits the local variables are cleared in the outer scope as well. This results in incorrect "variable not defined" runtime exceptions when a variable defined before the call statement are accessed after it.

### VCL code that cause the problem

*main.vcl*
```vcl
// Accessing a local variable from the calling scope.
sub access_var_test {
        set req.http.test = var.test;
}

sub test1_recv {
        declare local var.test STRING;
        set var.test = "incorrect";
        set req.http.test = "correct";

        call access_var_test;
}

sub blank {}

sub test2_recv {
        declare local var.test STRING;
        set var.test = "test";

        call blank;

        set req.http.test = var.test;
}
```

*main.test.vcl*
```vcl
// Test should fail at the `call_subroutine` with an runtime exception but instead fails
// at the assertion due the called function being able to read values in the calling subroutines
// scope.
sub test_access_outer_scope_local {
        testing.call_subroutine("test1_recv");
        assert.equal(req.http.test, "correct");
}

// Test should pass, instead fails with an undefined variable runtime error.
sub test_locals_reset_after_call {
        testing.call_subroutine("test2_recv");
        assert.equal(req.http.test, "test");
}
```

```
rgm@penguin:~/src/github.com/richardmarshall/falco$ git rev-parse --short HEAD
3fb1291
rgm@penguin:~/src/github.com/richardmarshall/falco$ make linux_amd64 && ./dist/falco test test_cases/local_vars/main.vcl 
GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build \
                 -ldflags "-X main.version=dev" \
cd ./dist/ && cp ./falco-linux-amd64 ./falco && tar cfz falco-linux-amd64.tar.gz ./falco
Running tests... Done.
 FAIL  /home/rgm/src/github.com/richardmarshall/falco/test_cases/local_vars/main.test.vcl
  ●  [RECV] test_access_outer_scope_local

    Assertion Error: Assertion error: expect=correct, actual=incorrect
    Actual Value: incorrect

    5|         testing.call_subroutine("test1_recv");
    6|         assert.equal(req.http.test, "correct");
    7| }

  ●  [RECV] test_locals_reset_after_call

    Undefined variable var.test

    10| sub test_locals_reset_after_call {
    11|         testing.call_subroutine("test2_recv");
    12|         assert.equal(req.http.test, "test");

0 passed, 2 failed, 2 total, 1 assertions
```

### Expected behavior

* Accessing local variables defined in a calling subroutine should fail.
* Local variables in the calling subroutine should be retained after the completion of a call statement.

richardmarshall:
Save local variable state / regex match at start of `ProcessSubroutine` and restore them when returning.

Also includes adding call count tracking of function subroutines.

Resolves #253

### DIFF
--- a/interpreter/subroutine.go
+++ b/interpreter/subroutine.go
@@ -22,10 +22,16 @@ func (i *Interpreter) ProcessTestSubroutine(scope context.Scope, sub *ast.Subrou
 
 func (i *Interpreter) ProcessSubroutine(sub *ast.SubroutineDeclaration, ds DebugState) (State, error) {
 	i.process.Flows = append(i.process.Flows, process.NewFlow(i.ctx, sub))
-	// reset all local values and regex capture values
+
+	// Store the current values and restore after subroutine has ended
+	regex := i.ctx.RegexMatchedValues
+	local := i.localVars
+	i.ctx.RegexMatchedValues = make(map[string]*value.String)
+	i.localVars = variable.LocalVariables{}
+
 	defer func() {
-		i.ctx.RegexMatchedValues = make(map[string]*value.String)
-		i.localVars = variable.LocalVariables{}
+		i.ctx.RegexMatchedValues = regex
+		i.localVars = local
 		i.ctx.SubroutineCalls[sub.Name.Value]++
 	}()
 
@@ -57,6 +63,7 @@ func (i *Interpreter) ProcessFunctionSubroutine(sub *ast.SubroutineDeclaration, 
 	defer func() {
 		i.ctx.RegexMatchedValues = regex
 		i.localVars = local
+		i.ctx.SubroutineCalls[sub.Name.Value]++
 	}()
 
 	var err error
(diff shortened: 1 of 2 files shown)


ysugimoto (approved):
Thanks, this is the exact behavior of subroutine.
I'm silly that I only implemented in functional subroutine https://github.com/ysugimoto/falco/pull/254/files#diff-0cea9fc2db654689e70a287bccd11b7cf7d27456e086bd803692b8678eb9780fR57-R61 😓

PR #256, opened 2024-02-11
Handle % string escapes

### LINKED ISSUE #246: [Feature] Support percent escapes in string literals
### Kind of proposals

You can turn on multiple checkboxes.

- Parser
- [x] Linter
- Simulator
- Testing

### Describe the feature

Yesterday I had an issue with broken VCL, when upon upload Fastly was spitting following error:
`Unexpected byte 0xAD in UTF-8 in _short-string_`

I spent quite a bit of time to figure that "%" symbols in a string literal are not allowed and the linter didn't detect this

**required**
Please write solve problem when the feature is added

_Put example VCL after the feature has been implemented_

`if (req.url.path ~ "%test%")`
should output an error 

Here is the docs page which help me to figure where I was wrong:
https://developer.fastly.com/reference/vcl/regex/

richardmarshall:
Currently we are parsing string escapes using the more common `\` style. VCL string escapes use a different format and are considerably more complicated.

The three forms of escapes supported by VCL are UTF-8 byte escapes `%XX`, fixed width unicode code point escapes `%uXXXX` and variable width unicode code point escapes `%u{...}`.

The utf-8 escapes represent bytes of a encoded escape and can be between 1 and 4 sequential escapes. The rules for utf-8 encoding are enforced and the decoded code point must be a valid. Arbitrary byte sequences of binary data is not supported.

The two forms of unicode code point escapes are differentiated by the fact that the fixed width escapes can only represent code points up to 0xFFFF while the variable width escapes support up to 0x10FFFF.

For all escape types a zero value is treated as a terminator for the string and the remaining portion of the string is dropped.

More information here:
https://developer.fastly.com/reference/vcl/types/string/

This PR removes the `\` handling logic and adds support for the 3 `%` escape types.

Additionally due to the fact that the quorum field for directors is represented as a percentage with a `%` the handling of these percentage expressions is updated as well to not be treated as a string. This also fixes a compatibility issue due to the fact that Fastly treats these percentages as a postfix expression.

Meaning code like this is valid:
```
    .quorum = 50 /* comment */ %;
```

I implemented this type of expression with a generic postfix expression handler for the `%` operator. After finishing that I think maybe it's unnecessary to support the general case postfix expression as the quorum percentage is as far as I can tell the only use case in VCL. I've left the general case implementation in place for now but let me know if you think that simplifying this and only supporting the percentage expression is the right way to go.

Resolves #246


ysugimoto:
@richardmarshall Thanks, but recently I haven't had enough time to review it. will do it later this weekend 🙏

interpreter/interpreter_test.go
```diff
@@ -90,8 +90,11 @@ func TestProcessDeclarations(t *testing.T) {
 			DirectorType: &ast.Ident{Value: "client"},
 			Properties: []ast.Expression{
 				&ast.DirectorProperty{
-					Key:   &ast.Ident{Value: "quorum"},
-					Value: &ast.String{Value: "20%"},
+					Key: &ast.Ident{Value: "quorum"},
```
ysugimoto, line 93: need `go fmt` probably

ysugimoto (approved):
Awesome implementation, your approach makes us more robust for string dealing.
and, postfix expression is also nice, we can implement it easier if we need more prefix units.
I need to reflect this change on https://github.com/ysugimoto/falco/pull/248 but I will do 👍 
I put a tiny nit but we can merge.

interpreter/interpreter_test.go
richardmarshall: Since the `Value` literal is now on multiple lines the alignment of `Key` with it was actually removed by `go fmt`.

ysugimoto:
@richardmarshall Could you resolve conflicts that cause by another PRs? thanks

richardmarshall:
> @richardmarshall Could you resolve conflicts that cause by another PRs? thanks

Done. 👍

PR #257, opened 2024-02-14
req.url behavior changed to represent raw url path

akrainiouk:
Fastly implementation treats req.url as a raw value while falco automatically urldecodes it
This PR implements consistent req.url behavior with fastly.
It also fixes presentation of escaped string values in the debugger

### DIFF
--- a/interpreter/function/builtin/urldecode.go
+++ b/interpreter/function/builtin/urldecode.go
@@ -3,12 +3,10 @@
 package builtin
 
 import (
-	"net/url"
-	"strings"
-
 	"github.com/ysugimoto/falco/interpreter/context"
 	"github.com/ysugimoto/falco/interpreter/function/errors"
 	"github.com/ysugimoto/falco/interpreter/value"
+	"net/url"
 )
 
 const Urldecode_Name = "urldecode"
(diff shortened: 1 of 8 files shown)


akrainiouk:
Several, related issues fixed in this PR:
* _req.url_ (and _req.url.path_) in Fastly implementation provides raw version of url and url path while an interpreter returns url decoded versions. This PR makes interpreter behavior consistent with Fastly.
* Also _urldecode_ implementation essentially applies url decoding twice, once by applying _url.PathUnescape_ and then by calling _url.QueryUnescape_. In %25 get yet another extra decoding. Fastly implementation is equivalent to a single application _url.QueryUnescape_. The confusion is likely caused by the fact that % escape is not decoded in quoted strings which is apparently fixed in [PR-256](https://github.com/ysugimoto/falco/pull/256)
* Fixed % escaped string representation in the debugger
* _testing.call_subroutine_ does not report an error when called with nonexistent function. The call is simply ignored.

interpreter/function/builtin/urldecode.go
```diff
@@ -38,18 +36,9 @@ func Urldecode(ctx *context.Context, args ...value.Value) (value.Value, error) {
...
 
-	dec, err := url.PathUnescape(input)
-	if err != nil {
-		return &value.String{IsNotSet: true}, errors.New(Urldecode_Name,
-			"Failed to urldecode string: %s", input,
-		)
-	}
-	// url.PathUnescape does not decode "+" sign to white space so we also need to call url.QueryUnescape
-	// in order to decode "+" sign into white space
-	dec, err = url.QueryUnescape(dec)
```
akrainiouk, line 52: In reality this only leads to double and triple decoding.

tester/function/testing_call_subroutine.go
```diff
@@ -49,6 +49,8 @@ func Testing_call_subroutine(
 	} else if sub, ok := ctx.Subroutines[name]; ok {
 		state, err = i.ProcessSubroutine(sub, interpreter.DebugPass)
 		i.TestingState = state
+	} else {
+		return value.Null, errors.NewTestingError("subroutine %s not found in VCL", name)
```
akrainiouk, line 53: If neither _SubroutineFunction_ not _Subroutine_ was found, report an error.

interpreter/function/builtin/urldecode_test.go
```diff
@@ -20,7 +20,7 @@ func Test_Urldecode(t *testing.T) {
 		expect string
 	}{
 		{input: "hello%20world+!", expect: "hello world !"},
-		{input: "hello%2520world+!", expect: "hello world !"},
+		{input: "hello%2520world+!", expect: "hello%20world !"},
```
akrainiouk, line 23: [Fastly documentation](https://developer.fastly.com/reference/vcl/functions/strings/urldecode/) is a bit misleading:
> For example, urldecode({"hello%20world+!"}); and urldecode("hello%2520world+!"); will both return "hello world !".

What it really means to demonstrate is that two different string literals `"..."` and `{"..."}` are interpreted differently, rather than double encoded space (%2520) is decoded to a space character.

akrainiouk:
NOTE: I am splitting this draft into smaller PRs so they are easier to handle

PR #261, opened 2024-02-16
fixed double decoding in urldecode

akrainiouk:
Current falco implementation of `urldecode` essentially applies url decoding twice, once by applying `url.PathUnescape` and then by calling `url.QueryUnescape`. In addition `%25` (escaped "%") gets yet another extra decoding.

[Fastly documentation](https://developer.fastly.com/reference/vcl/functions/strings/urldecode/) of urldecode is a bit misleading:

>    For example, urldecode({"hello%20world+!"}); and urldecode("hello%2520world+!"); will both return "hello world !".

This makes an impression that both %20 and %2520 should be decoded to space character
However in reality it means to demonstrate is that two different string literals "..." and {"..."} are interpreted differently. In the second example %25 is decoded to '%' by the paraser leading to the same original value passed to urldecode as in the first example.

The confusion is also likely caused by misinterpreting of quoted literals in falco which is apparently fixed in [PR-256](https://github.com/ysugimoto/falco/pull/256)

### DIFF
--- a/interpreter/function/builtin/urldecode.go
+++ b/interpreter/function/builtin/urldecode.go
@@ -3,12 +3,10 @@
 package builtin
 
 import (
-	"net/url"
-	"strings"
-
 	"github.com/ysugimoto/falco/interpreter/context"
 	"github.com/ysugimoto/falco/interpreter/function/errors"
 	"github.com/ysugimoto/falco/interpreter/value"
+	"net/url"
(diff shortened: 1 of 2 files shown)


richardmarshall (approved):
Thanks, this looks good.

ysugimoto (approved):

PR #262, opened 2024-02-17
req.url: fixed consistency with Fastly implementation

akrainiouk:
Fastly implementation  of req.url, req.url.path, req.url.basename and req.url.ext returns the raw version of these values while falco returns urldecoded versions.
This PR fixes this inconsistency

### DIFF
--- a/interpreter/variable/all.go
+++ b/interpreter/variable/all.go
@@ -39,6 +39,17 @@ func NewAllScopeVariables(ctx *context.Context) *AllScopeVariables {
 	}
 }
 
+// return unescaped path value from the specified URL
+// this function makes the best effort to get the raw path
+// even when standard getEscapedPath() chooses escaped version
+func getRawUrlPath(url *url.URL) string {
+	result := url.EscapedPath()
+	if url.RawPath != "" && result != url.RawPath {
+		result = url.RawPath
+	}
+	return result
+}
+
 // nolint: funlen,gocognit,gocyclo
 func (v *AllScopeVariables) Get(s context.Scope, name string) (value.Value, error) {
 	req := v.ctx.Request
@@ -499,7 +510,7 @@ func (v *AllScopeVariables) Get(s context.Scope, name string) (value.Value, erro
 		}
 		return &value.String{Value: u}, nil
 	case REQ_URL:
-		u := req.URL.EscapedPath()
+		u := getRawUrlPath(req.URL)
 		if v := req.URL.RawQuery; v != "" {
 			u += "?" + v
 		}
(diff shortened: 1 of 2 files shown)


richardmarshall:
Could you please provide an example of how Falco deviates from fastly on this? A Fiddle vs a falco vcl test case for example.

akrainiouk:
This fiddle demonstrates fastly behavior: https://fiddle.fastly.dev/fiddle/86d69ef4
Request path value contains url encoded characters as well as literal '+' character: `/d%69rname/ba%73e+name.extensi%6Fn`
the code in vcl_error dumps the values of various req.url subfields into the response body as a plain text.
Here is the response:
```

    req.url:          '/d%69rname/ba%73e+name.extensi%6Fn'
    req.url.path:     '/d%69rname/ba%73e+name.extensi%6Fn'
    req.url.dirname:  '/d%69rname'
    req.url.basename: 'ba%73e+name.extensi%6Fn'
    req.url.ext:      'extensi%6Fn'
```
As you can see all req.url fields maintains the original encoding.
It is a little tricker to see falco's behavior as I couldn't find the way to see the synthetic responses in the simulator.
The easiest way is probably to save this fragment into the file (e.g. main.vcl):
```
sub vcl_recv {
  // @debugger
  return(lookup);
}
```
and run it in the debugger:
```bash
falco simulate -debug main.vcl
```
Send test http request:
```bash
curl -Ss 'http://localhost:3124/d%69rname/ba%73e+name.extensi%6Fn'
```
And then when the execution breaks into debugger just manually url components while watching the output:
```
> req.url
>> (STRING)/dirname/base+name.extension
> req.url.path
>> (STRING)/dirname/base+name.extension
> req.url.dirname
>> (STRING)/dirname
> req.url.basename
>> (STRING)base+name.extension
> req.url.ext
>> (STRING)extension
```
Node that as opposed to Fastly behavior all url components come out decoded.
'+' character is (correctly) preserved though.

richardmarshall:
Thank you for the details. So this is an interesting case; these changes do fix a consistency issue but also break an edge case where we're currently compatible with Fastly.

Fastly allows you to set `req.url` to a value that is not escaped properly and will not escape the value when read back.

```vcl
set req.url = "/a b";
log req.url; // yields `/a b` not `/a%20b`
```

Granted we're only compatible with this edge case by way of an accident but still worth maintaining. It should be pretty straightforward to retain compatibility by comparing the result of `EscapedPath()` against `RawPath`. If `RawPath` is not empty and is not equal to the escaped value return the `RawPath` instead of the escaped path.

Something along the lines of:
```go
u := req.URL.EscapedPath()
if req.URL.RawPath != "" && u != req.url.RawPath {
    u = req.URL.RawPath
}
```

akrainiouk:
@richardmarshall thanks for your advice. I amended code accessing url raw path as you suggested and this took care of handling of spaces.
One case that is not covered is when assigned URL value contains invalid URL encoding.
In Falco it leads to parsing error (because it attempts to pre-parse the value into URL object).
Fastly apparently just holds req.url as a string and only splits it into path ? query # fragment when necessary.

Fixing this in Falco will require deeper changes so I am holding for now.

interpreter/variable/all.go
```diff
@@ -39,6 +39,17 @@ func NewAllScopeVariables(ctx *context.Context) *AllScopeVariables {
 	}
 }
 
+// return unescaped path value from the specified URL
+// this function makes the best effort to get the raw path
+// even when standard getEscapedPath() chooses escaped version
```
richardmarshall, line 44: Typo, `EscapedPath()`

interpreter/variable/all.go
```diff
@@ -39,6 +39,17 @@ func NewAllScopeVariables(ctx *context.Context) *AllScopeVariables {
...
+func getRawUrlPath(url *url.URL) string {
```
richardmarshall, line 45: Please change this argument name.

richardmarshall (changes requested):
Overall looks good. Just one typo and a linter issue that need fixing.

Regarding the invalid escape sequence issue I agree that it can be dealt with separately.

interpreter/variable/all.go
akrainiouk: Thanks! Goes to show golang is not my primary language :)
Fixed.

interpreter/variable/all.go
akrainiouk: Fixed a well, sorry for missing this

akrainiouk:
> Regarding the invalid escape sequence issue I agree that it can be dealt with separately.

There is a similarity in behavior of req.url and querystring function family (see #264).

From what I can see from Fastly behavior req.url handling shares the same implementation as querystring functions, at least in the sense that they both avoid full URL parsing. Perhaps the same code can be used for both.

richardmarshall (approved):

ysugimoto (approved):

PR #263, opened 2024-02-18
fix base64 decode related builtin function

### LINKED ISSUE #255: [BUG] Base64 decode process is different from the actual one
### Kind of proposals
- [ ] Parser
- [ ] Linter
- [x] Simulator
- [x] Testing

### Describe the problem
The processing of these functions are different from fiddle.
digest.base64_decode
digest.base64url_decode
digest.base64url_nopad_decode

* If there are characters that cannot be processed during base64 decoding, the VCL will truncate them there instead of reporting an error.
    * falco currently shows `illegal base64 data at input byte <NUM>`  errors
* If a null character is included, the string ends there.
    * This is the behavior of String, not just base64
* In digest_base64url_nopad_decode, the `=` is first excluded from the entire text.

https://fiddle.fastly.dev/fiddle/a3b006c4

Here is my test case.
https://github.com/bungoume/falco-vcl-empty-test/blob/main/tests/base64_decode.test.vcl

### Expected behavior
The following is a general idea of the modification.

https://github.com/bungoume/falco/commit/a806d40b758d9da2d8098f2a6105836ba0238f3f
(I'm wondering where to write common processes and if this error handling approach is appropriate. So, just to sharing the diff URL)

 - OS: mac
 - Shell: zsh

ysugimoto:
Fixes #255

Fastly terminates null-byte on a decoded string in base64 decoding functions:

- `digest.base64_decode`
- `digest.base64url_decode`
- `digest.base64url_decode_nopad`

Additionally, `digest.base64url_decode_nopad` also stops decoding when a padding sign (`=`) is found.
This PR follows Fastly behavior that describes the [documentation](https://www.fastly.com/documentation/reference/vcl/functions/cryptographic/digest-base64-decode/#padding).

### DIFF
--- a/interpreter/function/builtin/digest_base64_decode.go
+++ b/interpreter/function/builtin/digest_base64_decode.go
@@ -52,10 +52,8 @@ var nullByte = []byte{0}
 
 // Stop and return the point of Null-Byte found
 func terminateNullByte(decoded []byte) []byte {
-	if found := bytes.Index(decoded, nullByte); found != -1 {
-		return decoded[:found]
-	}
-	return decoded
+	before, _, _ := bytes.Cut(decoded, nullByte)
+	return before
 }
 
 // Even stopping decoding, we need to padding sign to success golang base64 decoding
--- a/interpreter/function/builtin/digest_base64_decode.go
+++ b/interpreter/function/builtin/digest_base64_decode.go
@@ -87,10 +87,9 @@ func Digest_base64_decode_removeInvalidCharacters(input string) string {
 		case b == 0x3D: // =
 			// If "=" sign found, next byte must also be "="
 			if peek, err := r.Peek(1); err != nil && peek[0] == 0x3D {
-				// If removed bytes is not multiple of 4bytes, the sign is padding
 				removed.WriteByte(b)
 				removed.WriteByte(b)
-				r.ReadByte() // skip next equal sign
+				r.ReadByte() // skip next "=" character
 				continue
 			}
 			// Otherwise, treat as invalid character, stop decoding
--- a/interpreter/function/builtin/digest_base64_decode.go
+++ b/interpreter/function/builtin/digest_base64_decode.go
@@ -3,6 +3,7 @@
 package builtin
 
 import (
+	"bytes"
 	"encoding/base64"
 
 	"github.com/ysugimoto/falco/interpreter/context"
@@ -37,10 +38,16 @@ func Digest_base64_decode(ctx *context.Context, args ...value.Value) (value.Valu
 	}
 
 	input := value.Unwrap[*value.String](args[0])
-	dec, err := base64.StdEncoding.DecodeString(input.Value)
-	if err != nil {
-		return value.Null, err
-	}
+	dec, _ := base64.StdEncoding.DecodeString(input.Value)
 
-	return &value.String{Value: string(dec)}, nil
+	return &value.String{Value: string(terminateNullByte(dec))}, nil
+}
+
+var nullByte = []byte{0}
+
+func terminateNullByte(decoded []byte) []byte {
+	if found := bytes.Index(decoded, nullByte); found != -1 {
+		return decoded[:found]
+	}
+	return decoded
 }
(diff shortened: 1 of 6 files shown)


richardmarshall (changes requested):
Proper handling of null bytes in the decode output and ignoring decoding errors definitely gets us closer to matching Fastly's behavior but there's still one issue.

> Invalid characters are skipped entirely when decoding, and do not contribute to the bits for the current byte being output. Decoding continues for subsequent data as if the invalid character were not present.

Since the go decoder stops processing the input on invalid characters they should first be removed from the input before passing to the decoder.

ysugimoto:
I looked into Fastly documentation and improved base64-ed string dealing with corresponding functions.
Adding more edge test cases and be sure Faslty's actual behavior: https://fiddle.fastly.dev/fiddle/b2e84dcd

interpreter/function/builtin/digest_base64_decode.go
```diff
@@ -37,10 +41,63 @@ func Digest_base64_decode(ctx *context.Context, args ...value.Value) (value.Valu
...
+// Base64 decoding utility functions - they are also called by digest.base64url_decode and digest.base64_decode_nopad
+var nullByte = []byte{0}
+
+// Stop and return the point of Null-Byte found
+func terminateNullByte(decoded []byte) []byte {
+	if found := bytes.Index(decoded, nullByte); found != -1 {
+		return decoded[:found]
+	}
+	return decoded
+}
```
richardmarshall, line 59: This implementation works and this isn't important but the `strings.Cut` function would be a 1 line option for this that doesn't need any nesting.

```go
func terminateNullByte(decoded string) string {
    decoded, _, _ = strings.Cut(decoded, "\x00");
    return decoded;
}
```

interpreter/function/builtin/digest_base64_decode.go
```diff
@@ -37,10 +41,63 @@ func Digest_base64_decode(ctx *context.Context, args ...value.Value) (value.Valu
...
+		b = append(b, 0x3D)
+	}
+	return b
+}
+
+func Digest_base64_decode_removeInvalidCharacters(input string) string {
+	removed := new(bytes.Buffer)
+	r := bufio.NewReader(strings.NewReader(input))
+
+	for {
```
richardmarshall, line 73: This question applies for the other `*removeInvalidCharacters` functions. Is there a specific reason why you went with the `bufio.Reader` approach here instead of ranging over the `input` string?

richardmarshall (commented):
Overall looks good, just a couple questions on the diff.

Sorry for not getting to this sooner I briefly looked at it when you submitted it and didn't have time to do a real review and forgot to leave the notification marked as unread. Oops.

interpreter/function/builtin/digest_base64_decode.go
ysugimoto: I intend to use peek bytes to check the next byte without string range checking.

interpreter/function/builtin/digest_base64url_decode_test.go
```diff
@@ -14,20 +14,76 @@ import (
...
+// 	}
+// 	if ret.Type() != value.StringType {
+// 		t.Errorf("Unexpected return type, expect=STRING, got=%s", ret.Type())
+// 	}
+// 	v := value.Unwrap[*value.String](ret)
+// 	expect := "some data with "
+// 	if v.Value != expect {
+// 		t.Errorf("return value unmach, expect=%s, got=%s", expect, v.Value)
+// 	}
+// }
```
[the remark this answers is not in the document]
ysugimoto: Ah this is unnecessary, this case is covered in the above new test table.

interpreter/function/builtin/digest_base64_decode.go
ysugimoto: I will use `bytes.Cut` for it, thanks.

richardmarshall (approved):

richardmarshall:
LGTM

PR #266, opened 2024-02-26
Fixed parsing of += operator

akrainiouk:
As it turns out "+=" operator in **set** statement is parsed to "=" as a result corresponding
code is also misinterpreted as simple assignment:
```varnish-vcl
  declare local var.number INTEGER; 
  set var.number = 3;
  set var.number += 2;
```
results in var.number value equal to 2 rather then 5

### DIFF
--- a/parser/statement_parser_test.go
+++ b/parser/statement_parser_test.go
@@ -1,6 +1,7 @@
 package parser
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/ysugimoto/falco/ast"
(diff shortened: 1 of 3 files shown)


parser/statement_parser_test.go
```diff
@@ -77,47 +78,58 @@ include "feature_mod" // Trailing comment`
...
-		vcl, err := New(lexer.NewFromString(input)).ParseVCL()
-		if err != nil {
-			t.Errorf("%+v", err)
+			}
+			vcl, err := New(lexer.NewFromString(input)).ParseVCL()
+			if err != nil {
+				t.Errorf("%+v", err)
+			}
+			assert(t, vcl, expect)
+
```
akrainiouk, line 131: This is copy of the original _simple assign_ test that is extended to verify all assignment operators.

richardmarshall (approved):
Nice catch, thanks. LGTM.

PR #269, opened 2024-03-09
keepalive_time added to supported backend properties

akrainiouk:
I don't see this property documented but support for it was added about a year ago:
https://www.fastly.com/documentation/reference/changes/2023/02/keepaplive-time-is-seconds/
I also added this one as well as other valid properties to backend statement 'pass' test.


ysugimoto (approved):
Looks good, thanks.

PR #270, opened 2024-03-26
improve trailing/infix comment parsing

ysugimoto:
This PR fixes a tiny bug about parsing comment tokens, particularly `InfixComment` and `TrailingComment`.
Please look at the following spec.

### Comment Parsing

From the VCL parsing spec, we always have the next token as `peekToken`, including the leading comment but the leading comment will have the InfixComment or Trailing Comment for the `currentToken`. This causes problems that:

```vcl
sub vcl_recv {
    set req.http.Foo = "bar";
    // this comment should be a InfixComment in BlockStatement of subroutine  - (1)
} // this comment should be a TrailingComment in BlockStatement of subroutine -(2)
// this comment should be a LeadingComment in the next token - (3)
```

The previous implementation had a problem in that (2) and (3) comments could not parse correctly because both comments have a leading comment in `peekToken`.

### Solving

I changed the comment parsing implementation by defining the `PrefixedLineFeed` property in `ast.Comment`, this property indicates that the previous token of comment had an `LF` or not, and by checking whether this property is true, we can complete distinguishing `TrailingCommang` and `LeadingComment`.

This implementation won't affect to main features like lint, and interpreter, just for comment treating but needed to implement formatted (now I'm working on it).

### DIFF
--- a/parser/parser_test.go
+++ b/parser/parser_test.go
@@ -196,7 +196,7 @@ sub vcl_recv {
 func TestAllCommentPositions(t *testing.T) {
 	input := `
 // subroutine leading
-sub /* subroutine ident leading */ vcl_recv {
+sub /* subroutine ident leading */ vcl_recv /* subroutine block leading */ {
 	// if leading
 	if (
 		req.http.Host &&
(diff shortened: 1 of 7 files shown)


parser/parser_test.go
```diff
@@ -192,3 +192,86 @@ sub vcl_recv {
 	}
 	assert(t, vcl, expect)
 }
+
+func TestAllCommentPositions(t *testing.T) {
+	input := `
+// subroutine leading
+sub /* subroutine ident leading */ vcl_recv {
```
richardmarshall, line 199: For completeness it would be good to include a subroutine block leading comment.

```
sub /* subroutine ident leading */ vcl_recv /* subroutine block leading */ {
```

Also maybe before / after the parens of the if statement condition.

richardmarshall (commented):
Overall it looks good. Just a comment about test completeness.

A formatting command will be a nice addition. Do you have a document that could be shared covering the planned formatting rules?

ysugimoto:
Okay, it's easy to handle it, fixed.

parser/parser_test.go
ysugimoto: Okay, it's easy to handle it - done.

richardmarshall (approved):

ysugimoto:
I will share the current my considerations about the formatting rules later.

PR #275, opened 2024-03-29
special dealing for req.hash addition assignment

ysugimoto:
This PR fixes the edge case bug - normally `+=` operator cannot be used for STRING but can be used for `req.hash` predefined variable for adding request hash.

## Normal case (fail)

```vcl
sub vcl_recv {
  set req.http.Foo += "foo"; // error
}
```

## Edge case (should pass)

```vcl
sub vcl_hash {
  set req.hash += "foo"; // pass
}
```

## Reproduce

```vcl
// test.vcl
sub vcl_hash {
  set req.hash += req.url;
}
```

```
🔥 [ERROR] Operator "+=" cannot be used for STRING (operator/conditional)
in /Users/sugimoto/github/falco/cmd/falco/test.vcl at line 2, position 16
 1|sub vcl_hash {
 2|  set req.hash += req.url;
                  ^^
 3|}
See reference documentation: https://developer.fastly.com/reference/vcl/operators/#conditional-operators
```

Fastly fiddle: https://fiddle.fastly.dev/fiddle/0f3fc0aa

### DIFF
--- a/interpreter/variable/hash.go
+++ b/interpreter/variable/hash.go
@@ -69,6 +69,14 @@ func (v *HashScopeVariables) Get(s context.Scope, name string) (value.Value, err
 
 func (v *HashScopeVariables) Set(s context.Scope, name, operator string, val value.Value) error {
 	if name == "req.hash" {
+		// Special string assignment - normally "+=" operator cannot use for STRING type,
+		// But the exception case that "+=" operation can use for the "req.hash".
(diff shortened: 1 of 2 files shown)


richardmarshall (changes requested):
This change to the linter look good but the interpreter will also need to be updated to handle this case correctly.

richardmarshall (approved):
LGTM

PR #277, opened 2024-03-30
Prioritize cache object

### LINKED ISSUE #276: [BUG] Incorrect simulator response when request transitions from FETCH -> ERROR
### Kind of proposals

- [x] Simulator

### Describe the problem

When running the simulator if a request transitions from FETCH to ERROR when the request is then processed in DELIVER the resp object will contain the backend response data instead of the  `obj` values setup in the ERROR state.

```vcl
backend glitch {
  .host = "http-me.glitch.me";
  .port = "443";
  .ssl = true;
}

sub vcl_recv {
  set req.backend = glitch;
  set req.url = "/status=200";
}

sub vcl_fetch {
  error 418;
}

sub vcl_deliver {
  log resp.status;
}
```

Log statement output in `vcl_deliver` shows 200 instead of 418.
```shell
falco simulate main.vcl                              
Simulator server starts on 0.0.0.0:3124
Request Incoming =========>
Move state: RECV -> HASH
Move state: HASH -> MISS
Fetching backend (glitch) https://http-me.glitch.me:443/status=200
Move state: MISS -> FETCH
Backend (glitch) responds status code 200
Move state: FETCH -> ERROR
Move state: ERROR -> DELIVER
200
Move state: DELIVER -> LOG
<========= Request finished
```

Simulator response shows a `.client_response.status_code` of 200 instead of 418.
```shell
curl -s http://localhost:3124 | jq '.client_response.status_code'
200
```

### Expected behavior

The `resp` object in vcl_deliver after passing through vcl_error should contain the contents of `obj` not `bereq`.

https://fiddle.fastly.dev/fiddle/9a17cfc1

The simulator output `.client_response` should contain the values set in `vcl_error`.

**required**
 - OS: linux
 - Shell: zsh

richardmarshall:
Currently the setup of `i.ctx.Response` in `ProcessDeliver` is prioritizing `i.ctx.BackendResponse` if it is set. This causes a problem for requests that transition with this flow: `FETCH -> ERROR -> DELIVER`. The response setup in `vcl_error` is lost and the simulator incorrectly uses the backend response.

Additionally the deferred closure in `ProcessFetch` is setting `i.ctx.Response` to be the backend response. This does not impact VCL code in `vcl_deliver` like the previous issue but it does impact the simulator output as the correct response data is overwritten when ProcessFetch returns which occurs after the request handling is complete but before the `i.process.Finalize` is called.

While fixing those issues also discovered that when the simulator moves to the error state the obj value is not always setup to be an empty response. Values from the backend response or cache object will be included in the response which is incorrect as described here: https://www.fastly.com/documentation/reference/vcl/subroutines/error/.

Fixes #276

### DIFF
--- a/interpreter/interpreter.go
+++ b/interpreter/interpreter.go
@@ -518,7 +518,7 @@ func (i *Interpreter) ProcessFetch() error {
 	}
 	// TODO: consider stale-white-revalidate and stale-if-error TTL
 
-	// Consider cache, create client response from backend response
+	// Update cache
 	defer func() {
 		resp := i.cloneResponse(i.ctx.BackendResponse)
 		// Note: compare BackendResponseCacheable value
@@ -533,7 +533,6 @@ func (i *Interpreter) ProcessFetch() error {
 				})
 			}
 		}
-		i.ctx.Response = resp
 	}()
 
 	// Simulate Fastly statement lifecycle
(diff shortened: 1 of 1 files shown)


ysugimoto (approved):
We should prioritize looking up Object rather than BackendResponse surely, nice catch.

PR #278, opened 2024-03-30
fix exprression comment parsing

ysugimoto:
This PR fixes more comment parsing logic in `parser`.

## Problem 

We have a problem parsing inline comments on the last of a token (before the semicolon) in `InfixExpression`:

```vcl
set vcl_recv {
  set req.http.Foo = /* comment 1 */ "bar" /* comment 2 */ "baz" /* comment 3 */;
}
```

The current parser lacks `/* comment 3 */` in `InfixExpression`, so this PR fixes this.

## Plan to solve

I'm focusing on what the `/* comment 2 */` comment  should own where `Left` or `Right` expression, I thought that both expressions should have it in trailing comment in `Left`:

```vcl
set vcl_recv {
  set req.http.Foo = /* comment 1 */ "bar" /* comment 2 */ "baz" /* comment 3 */;
  // Left:           ^ leading comment     ^ trailing comment
  // Right:                                                      ^ trailing comment
}
```

The `InfixExpression` node may have some comments but they should be ignored for formatting because Left and Right expressions can represent all comments to be formatted.

### DIFF
--- a/parser/expression_parser.go
+++ b/parser/expression_parser.go
@@ -73,12 +73,17 @@ func (p *Parser) parseExpression(precedence int) (ast.Expression, error) {
 		if !ok {
 			return left, nil
 		}
+		swapLeadingTrailing(p.peekToken, left.GetMeta())
 		p.nextToken()
 		left, err = infix(left)
 		if err != nil {
 			return nil, errors.WithStack(err)
 		}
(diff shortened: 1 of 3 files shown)


richardmarshall (approved):
LGTM

PR #282, opened 2024-03-30
Errors with concurrent simulator requests due to global interpreter state

richardmarshall:
Currently if the interpreter is not threadsafe as much of the state for an individual request being processed is stored in the interpreter struct itself. This causes problems for the simulator if it receives multiple concurrent requests.

While it might be nice for the interpreter to be threadsafe reworking things for that would be rather involved. A simpler solution to solve this specific issue is to add an interpreter lock which the simulator HTTP handler can acquire before processing a request.

This PR adds lock/unlock behavior for the interpreter http handler as well as setting up the Fastly-FF header to let the handler detect loops. Without that a deadlock could occur if the simulator was configured to be a backend for itself.

### DIFF
--- a/interpreter/handler.go
+++ b/interpreter/handler.go
@@ -7,14 +7,16 @@ import (
 	"github.com/pkg/errors"
 	"github.com/ysugimoto/falco/interpreter/exception"
 	"github.com/ysugimoto/falco/interpreter/limitations"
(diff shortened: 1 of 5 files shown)


ysugimoto (approved):
Looks good.

I agree with you because the simulator is just for local debugging, not for production use so it's okay to be locked by mutex.

PR #283, opened 2024-03-30
Fix offset/length handling in substr & utf8.substr

richardmarshall:
The offset and length handling logic in `substr` and `utf8.substr` currently result in incorrect results for some inputs.

For example:

```vcl
sub test {
  assert.equal(substr("abcdefg", 0, 2), "ab");
}
```

```shell
  ●  [RECV] test

    Assertion Error: Assertion error: expect=ab, actual=abc
    Actual Value: abc

    1| sub test {
    2|   assert.equal(substr("abcdefg", 0, 2), "ab");
    3| }

0 passed, 1 failed, 1 total, 1 assertions
```

The correct result is `ab` for a substring of length 2. Fiddle showing fastly handles this correctly https://fiddle.fastly.dev/fiddle/9de9a70b

This PR fixes this case as well as improving compatibility with Fastly behavior for some edge cases.
* If the calculated starting point of substr is negative return empty string
* Detect integer overflow and return empty string
* Offset beyond length of string returns empty string

### DIFF
--- a/interpreter/function/builtin/substr.go
+++ b/interpreter/function/builtin/substr.go
@@ -35,7 +35,7 @@ func Substr(ctx *context.Context, args ...value.Value) (value.Value, error) {
 		return value.Null, err
 	}
 
-	input := []byte(value.Unwrap[*value.String](args[0]).Value)
+	input := value.Unwrap[*value.String](args[0]).Value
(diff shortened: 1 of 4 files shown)


ysugimoto (approved):
Thanks for the fix, nice catch.

PR #284, opened 2024-03-31
Regex patterns must be literals

richardmarshall:
This PR fixes two issues, first that regex patterns must be string literals as they are handled at compile time.

https://www.fastly.com/documentation/reference/vcl/regex/
> Where an operator or function expects a regex, you must provide a literal pattern in your code.

To fix this in Falco the `Regex` operator function as well as the regex related builtin vcl functions needed to be updated to check that the pattern argument is a string literal. Also updated the linter to detect non-literal patterns for the regex match operators and `regsub*` function arguments.

Second there is a bug in `regsub` where non-matching portions of the input string can get lost.

For example the case:
```vcl
regsub("foo;bar", "^([^;]+)", "\1");
```

The result of this should be the same as the input string `foo:bar` because the regex pattern is matching the first segment of the input `foo` and replacing it with the match group 1 which is the same value. However the current `regsub` logic results in the non-matching portion of the input to be lost.

```vcl
sub test {
  assert.equal(regsub("foo;bar", "^([^;]+)", "\1"), "foo;bar");
}
```

```shell
    Assertion Error: Assertion error: expect=foo;bar, actual=foo
    Actual Value: foo

    1| sub test {
    2|   assert.equal(regsub("foo;bar", "^([^;]+)", "\1"), "foo;bar");
    3| }
```

Fiddle showing the correct behavior on fastly: https://fiddle.fastly.dev/fiddle/3c908f13

Resolved this by replacing the current logic with an implementation that uses `ReplaceAllStringFunc` in a way that behaves like a `ReplaceOneStringFunc` function. Additionally changed the logic for converting the VCL replacement group format to the go format to one that uses a regex replacement.

### DIFF
--- a/interpreter/function/builtin/regsub.go
+++ b/interpreter/function/builtin/regsub.go
@@ -10,9 +10,28 @@ import (
 	"github.com/ysugimoto/falco/interpreter/value"
 )
 
-const Regsub_Name = "regsub"
+const (
+	Regsub_Name = "regsub"
 
-var Regsub_ArgumentTypes = []value.Type{value.StringType, value.StringType, value.StringType}
+	regSubExpandReplace = "${$1}"
+)
+
+var (
+	Regsub_ArgumentTypes = []value.Type{value.StringType, value.StringType, value.StringType}
+
+	regsubExpandRE = regexp.MustCompile(`\\([0-9]+)`)
+)
+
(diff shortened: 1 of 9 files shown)


ysugimoto (approved):
LGTM

PR #285, opened 2024-03-31
Add missing backend.{name}.* variables

richardmarshall:
Adds a few missing `backend.{name}.*` variables.

New variables and their values:
* `backend.{name}.connections_open` - `0`
* `backend.{name}.connections_used` - `0`
* `backend.{name}.healthy` - `true`


ysugimoto (approved):
LGTM

PR #289, opened 2024-04-02
Additional details in simulator output

richardmarshall:
## Report selected backend in flows

Currently the simulator does not provide a way to see the selected backend at each recorded flow. This information is helpful when validating the behavior of backend selection in vcl_recv vs vcl_miss/vcl_pass.

## Optional output of response body

Add a new cli flag to the simulator which toggles the output of the response body in `.client_response`.

## fix missing `--version` flag

Help output includes the `--version` flag for displaying the Falco version but the flag was missing from the Config struct tags.


ysugimoto:
This feature is covered by `--proxy`, actual proxing option.

PR #294, opened 2024-04-03
Store function call state in a stack

### LINKED ISSUE #293: [BUG]
### Kind of proposals

- [x] Simulator
- [x] Testing

### Describe the problem

Currently if a recursion loop is created the interpreter will not detect this and end up panicking with a stack overflow error.

```vcl
sub f1 {
        call f2;
}

sub f2 {
        call f1;
}
```

```vcl
sub test {
        testing.call_subroutine("f1");
}
```

```
% falco test main.vcl 
Running tests...runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0xc03f1ea348 stack=[0xc03f1ea000, 0xc05f1ea000]
fatal error: stack overflow
...
```

### Expected behavior

The interpreter should return an error and not panic.

richardmarshall:
As functions are called store the current execution state for the function in a stack. This includes the local variables and re.group.{N} variables.

This stack is used to detect recursion and prevent an interpreter panic.

In the future this information could be used for including a VCL stack trace in error output or for accessing calling function state in the debugger.

Note: The current code structure makes accessing the re.group variables directly from the stack pointer a challenge. To keep things simple I retained the save / restore model we currently have for those. If there is consensus on the idea of using a stack frame model we can decide if it makes sense to restructure things to make access through the stack pointer possible.

Resolves #293, #297


ysugimoto (approved):
Looks good approach, LGTM

interpreter/subroutine.go
```diff
@@ -20,20 +20,47 @@ func (i *Interpreter) ProcessTestSubroutine(scope context.Scope, sub *ast.Subrou
...
+	return false
+}
+
+func (i *Interpreter) pushStackFrame(sub *ast.SubroutineDeclaration) {
+	sf := &StackFrame{
+		Locals:     variable.LocalVariables{},
+		Regex:      make(map[string]*value.String),
+		Subroutine: sub,
+	}
+	i.Stack = append(i.Stack, sf)
```
ysugimoto, line 38: Tentatively we can define the max call stack frames and raise an error if exceeded, probably around 100 would be enough but up to you.

richardmarshall: Yeah that seems reasonable. I did a test and generated a silly call chain of 200 functions in a fiddle and it didn't cause an error. Seems like the limiting factor isn't the call depth but the overall workspace memory usage which would eventually be exhausted.

ysugimoto:
@richardmarshall Sorry I forgot to merge this PR and treid to resolve conflicts but some testing errors are found due to the head branch changes.

I'm so happy if you could adjust to main branch, thanks.

PR #298, opened 2024-04-05
Handle implicit type conversion when calling functions

MasonM:
This fixes errors like `[regsub] Argument 1 expects STRING type but BACKEND provided` when calling builtin functions and passing a non-string for a string argument. Fastly handles that by implicitly converting the non-string argument to a string:
> These types all have implicit conversions to strings, such that their values may be used in contexts where a STRING value is necessary.
https://www.fastly.com/documentation/reference/vcl/types/

Fiddle: https://fiddle.fastly.dev/fiddle/2f830b56

I wasn't sure the cleanest way of implementing this. Originally, I was thinking of doing this conversion in `ProcessFunctionCallExpression()` and `ProcessFunctionCallStatement()` (the two places that call functions) like we're already doing for the `IsIdentArgument()` check: 
https://github.com/ysugimoto/falco/blob/65518ed8f04265e1a0ebe8e71d43122c10e6ec64/interpreter/statement.go#L418-L421

But it would need to know the argument types for every function to do that, which means encoding that information in the `Function` struct.

### DIFF
--- a/__generator__/interpreter.go
+++ b/__generator__/interpreter.go
@@ -124,7 +124,11 @@ func (i *Interpreter) generateBuiltInFunction() error {
 			return err
 		}
 		signature := fmt.Sprintf(
-			"Call: func(ctx *context.Context, args ...value.Value) (value.Value, error) { return builtin.%s(ctx, args...) }",
+			`Call: func(ctx *context.Context, args ...value.Value) (value.Value, error) {
+				doImplicitTypeConversion(args, %s)
+				return builtin.%s(ctx, args...)
+			}`,
+			argumentTypes(v.Arguments),
 			ucFirst(strings.ReplaceAll(key, ".", "_")),
 		)
 		buf.WriteString(signature + ",\n")
--- a/__generator__/template.go
+++ b/__generator__/template.go
@@ -85,6 +85,17 @@ import (
 	"github.com/ysugimoto/falco/interpreter/function/builtin"
 )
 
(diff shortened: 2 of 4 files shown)


richardmarshall (commented):
Funny I had actually gotten started on this problem due to the exact same thing you mention in the PR description, `regsub` with a backend value. But I ran a can of worms with edge cases and got distracted with other things and hadn't gotten to submitting a PR yet.

Largely we took the same general approach except I did the type coercion in the various function definition files themselves vs in `builtin_functions.go`. 

On my first look I initially considered your solution a better approach. However as I think more about the various edge case thorns I ran into when working on this I'm now not so sure. My main concern is that by having the coercion happen outside the function definition it separates the argument handling logic from the function implementation itself. This creates cases of unexpected "magic" happening to the arguments before the code in the function implementation sees it. Given the amount of edge cases there are I think it makes sense to have all the logic related to handling the arguments with the implementation of the functions.

For example we have variadic functions (`header.filter`, `early_hints`, etc), functions that require literal string arguments (`regsub*`, `header.filter`, etc), functions that require expressions of string literals (`querystring.filter*`), and I wouldn't be surprised if there are others that haven't been found yet.

By keeping the logic of type coercion with the other argument handling logic we make it easier to know where to look when trying to identify all of the details of how arguments are handled for a specific built-in function which is especially important for the edge case functions.

For reference this is the diff for my WIP branch: https://github.com/ysugimoto/falco/compare/main...richardmarshall:falco:regsub_backend_type

ysugimoto:
This is solved on #453 .

Issue #299, opened 2024-04-06
[BUG] Missed comment attachment during parsing


richardmarshall:
### Kind of proposals

- [x] Parser

### Describe the problem

While reviewing and testing out the new formatting functionality (#291) I ran into a number of issues related to comments. There are a lot of places where comments in the input VCL are lost in the parsing process and as a consequence are missing from the formatted output. While this is largely most relevant to the formatting efforts I'm making this a separate issue since at its root this is a parser problem.

To use an over the top example, for the following VCL input:
```vcl
/* 1 * / declare /* 2 * / local /* 3 * / var.s /* 4 * / STRING /* 5 * / ; /* 6 * /
```

In this case only comments 1 and 6 are captured as during parsing and a print of this node would yield
```vcl
/ * 1 * /
declare local var.s STRING; /* 6 * /
```

There are many other cases of valid comment positions that get lost during parsing, before semi-colons, trailing comments at the end of files, etc. Some of the loss is simply in the printing and the comments are attached to nodes. It's just the printing logic isn't emitting them, others are completely lost when tokens are read without a node being created for them.

### TL;DR

We're currently losing comments but our existing comment attachment points should be flexible enough to correctly capture all possible comments. To do so will require tweaks to the parser to ensure all of them are correctly attached to their relevant node. Will submit a PR with a starting point implementation for discussion.

### Comment attachment point analysis

Currently our AST node meta struct can handle a max of 3 comment attachment points (leading, infix, trailing) so it is important to confirm that this is enough to handle all cases. The following is an attempt to cover the VCL syntax and verify that our current model is sufficient to represent all comment positions as well as identifying what nodes and which positions each comment location should be attached to.

#### Position markers:

* `<start>` - comment immedietly preceding the node
* `<end>` - comment immediatly following the node
* `<after_keyword>` - comment immedietaly following a keyword
* `<after_bracket>` - comment immedietly following a bracket (`(`, `{`)
* `<x>` - comment is attached to a child node

#### Declarations:

**Untyped declarations:** (acl, table (implicit string), backend, ...)
```
<start> token <x> name <x> { <after_bracket>
  <x>
} <end>
```

attachment points: 3

**Typed declarations:** (table, sub)

```
<start> token <x> name <x> type <x> { <after_bracket>
  <x>
} <end>
```

attachment points: 3

#### Statements:

**Single token statements:** return (bare), break, fallthrough, esi, ...
```
<start> token <after_keyword> ; <end>
```

[...]

PR #302, opened 2024-04-19
Improve parser/ast for complex comments

ysugimoto:
This PR improves more complex comment parsing, which relates to #299 .

The current implementation focuses on the meaning syntax (IDENT, STRING, etc...) so some positions of comments are ignored. It's okay to run linter and interpreter but this is important for a formatter to keep the original VCL including comments.

See https://github.com/ysugimoto/falco/blob/feature/complex-comments/docs/parser.md which defines parser spec, and enables to put comment position.


ysugimoto:
I will merge this PR this weekend. If you have any opinions to improve, make another PR, please.

PR #330, opened 2024-07-13
Fix broken result message of tester and use exactly the same format for both passed and failed tests

nodaguti:
Hi, thank you very much for creating a great tool to help us work with VCL!

I've noticed that in result messages of tester,

1. `scope` is not placed in brackets `[]` of a failed test, which differs from passed one
2. `group` is not shown at all, while there is code to try to output it when a test fails

Screenshot 2024-07-13 at 10 12 40 AM

This PR fixes the above issues and lets `falco` report group name of tests even when a test passes for consistency.

Screenshot 2024-07-13 at 10 11 30 AM

### DIFF
--- a/cmd/falco/main.go
+++ b/cmd/falco/main.go
@@ -339,9 +339,9 @@ func runTest(runner *Runner, rslv resolver.Resolver) error {
 
 		for _, c := range r.Cases {
 			totalCount++
-      var prefix string
-				if c.Group != "" {
-					prefix = c.Group + " › "
+			var prefix string
+			if c.Group != "" {
+				prefix = c.Group + " › "
 			}
 			if c.Error != nil {
 				writeln(redBold, "%s● [%s] %s%s\n", indent(1), c.Scope, prefix, c.Name)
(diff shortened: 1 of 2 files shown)


cmd/falco/main.go
```diff
@@ -339,12 +339,12 @@ func runTest(runner *Runner, rslv resolver.Resolver) error {
 
 		for _, c := range r.Cases {
 			totalCount++
-			if c.Error != nil {
-				var prefix string
+      var prefix string
```
ysugimoto, line 342: Nits
Please run `go fmt`

tester/tester.go
```diff
@@ -223,6 +223,7 @@ func (t *Tester) runDescribedTests(
 			err := i.ProcessTestSubroutine(s, sub)
 			cases = append(cases, &TestCase{
 				Name:  suite,
+        Group: d.Name.String(),
```
ysugimoto, line 226: Nits
Please run `go fmt`

ysugimoto (approved):
Thanks, looks good with nits.

nodaguti:
@ysugimoto
Thank you for your review! I've addressed your comments.

PR #331, opened 2024-07-13
Add extra format check for backend.share_key

nodaguti:
According to fastly, `backend.share_key` must be alphanumeric and ASCII only as in an error message shown below returned from their server.

```
Syntax error: `.share_key` must be alphanumeric and ASCII only
at: (input Line 48 Pos 16)
  .share_key = "nodaguti-test.storage.googleapis.com";
---------------######################################-
```

Thus this PR introduces new check to linter so that `backend.share_key` will not contain any invalid characters.

### DIFF
--- a/linter/helper.go
+++ b/linter/helper.go
@@ -111,13 +111,13 @@ var DirectorPropertyTypes = map[string]DirectorProps{
 }
 
 func isAlphaNumeric(r rune) bool {
-	return (r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z') || (r >= '0' && r <= '9') || r == '_'
+	return (r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z') || (r >= '0' && r <= '9')
 }
 
 // isValidName validates ident name has only [0-9a-zA-Z_]+
 func isValidName(name string) bool {
 	for _, r := range name {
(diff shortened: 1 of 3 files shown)


linter/helper.go
```diff
@@ -111,13 +111,13 @@ var DirectorPropertyTypes = map[string]DirectorProps{
 }
 
 func isAlphaNumeric(r rune) bool {
-	return (r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z') || (r >= '0' && r <= '9') || r == '_'
+	return (r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z') || (r >= '0' && r <= '9')
```
nodaguti, line 114: The word `alphanumeric` usually refers to alphabets and numbers, and not underscores.

This is also the case in Fastly's documentation, for example

> Must start with an alphanumeric character and contain only alphanumeric characters, underscores, and whitespace.

at https://www.fastly.com/documentation/reference/api/acls/acl/ clearly indicates `alphanumeric characters` don't include underscores.

ysugimoto (approved):
Ah I wan not aware of it, thanks for your improvement.

nodaguti:
@ysugimoto
Thank you for your review! I've fixed a test whose assertion was flipped and run `go fmt'.

PR #334, opened 2024-07-16
fix(formatter): remove space between return and parenthesis

acme:
- Remove extra space before opening parenthesis in return statements
- Add simple test case
- Update test cases to reflect the new behaviour
- This matches the Fastly VCL style, e.g.:
   - https://www.fastly.com/documentation/reference/vcl/statements/return/ 
   - https://www.fastly.com/documentation/guides/vcl/using/#custom-vcl


ysugimoto (approved):
Nice catch. I agree with you for matching Fastly's style, thanks!

PR #335, opened 2024-07-16
feat(formatter): prevent indentation of #FASTLY macros

acme:
- Modified the `formatComment()` function to skip indentation for comments starting with `#FASTLY`
- Added unit tests to verify the new behaviour for regular comments, #FASTLY macros, and mixed comment scenarios
- This change ensures that #FASTLY macros remain at the start of the line, adhering to Fastly's VCL formatting conventions, e.g.:
  - https://www.fastly.com/documentation/guides/vcl/using/#custom-vcl


ysugimoto (approved):
I didn't care about this, thanks!

PR #342, opened 2024-07-25
fix(config): correct typo in SortDeclaration default tag

acme:
Fixes a typo in the `default` tag for the `SortDeclaration` field in the `FormatConfig` struct.


ysugimoto (approved):
Nice catch.

PR #349, opened 2024-09-23
feat: add `dap` subcommand and support basic DAP features

rinx:
This pull request adds `dap` subcommand and basic [Debug Adapter Protocol (DAP)](https://microsoft.github.io/debug-adapter-protocol/) features.

Falco already has a VCL debugger.
It's better to integrate DAP features with it, however the current debugger is highly coupled with TUI console.
Not to break compatibility, at the first step, I implemented DAP features as separated subcommand.

## Demo
In my Neovim environment, it works like below:
falco-dap-demo


ysugimoto:
@rinx
Thanks for your great implementation!
It seems to be enough for the begging of DAP support.
I will look into the implementation deeply, bear with my review.

dap/debugger.go
```diff
@@ -0,0 +1,117 @@
...
+
+	return d.mode
+}
+
+func (d *Debugger) getBreakpoint(node ast.Node) *breakpoint {
+	meta := node.GetMeta()
+
+	path, err := filepath.Abs(meta.Token.File)
+	if err != nil {
+		// TODO: handle error
```
ysugimoto, line 83: How should we treat for the error?
In terms of `ast.Node`, falco always treat as absolute path so `meta.Token.File` always has an absolute path.
Therefore you don't need to get the absolute path using `filepath.Abs`.

Additionally, falco may retrieves Fastly remote reources like `acl`, `backend`, etc then the remote source file will be `Remote.Acl.[name]` and the the getBreakpoint` works fine even not a filepath?

ysugimoto (changes requested):
Overall looks good, and put one comment for getting breakpoint.

[TODO] We need to write documentation for DAP usage, should be done in another PR.

dap/debugger.go
rinx: Thank you for your comment. I've just fixed it!

> Additionally, falco may retrieves Fastly remote reources like `acl`, `backend`, etc then the remote source file will be `Remote.Acl.[name]` and the the getBreakpoint` works fine even not a filepath?

Currently, it doesn't matter because `dap` subcommand does not support remote resources feature.
We need to implement it as a future work.

ysugimoto (approved):
Thanks! This should be merged.

PR #367, opened 2024-10-17
docs(dap): add a brief documentation about dap subcommand

rinx:
Refs: #349 

I wrote a brief documentation about `dap` subcommand.
I've just wrote about Neovim usecase.

BTW: I think it is good to support this feature in VSCode, however it requires to write some extension codes to run DAP client.
I found https://github.com/fastly/vscode-fastly-vcl/, but it seems not actively development. :thinking:


ysugimoto (approved):
Thanks, looks good.

PR #376, opened 2024-10-22
Fix Grouped Testing syntax

bungoume:
(before|after)_xxx with sub will result in an error. It would be correct not to include it?

```
Failed to run test: Parse Error: Unexpected token "before_recv", expects IDENT
```


ysugimoto (approved):
Ah yes that's correct, thanks.

PR #379, opened 2024-10-26
Add support for long strings with heredoc delimiters

gabrielg:
Fastly VCL allows [two kinds of "long strings"](https://www.fastly.com/documentation/reference/vcl/types/string/), one kind is simply enclosed within braces, e.g.

```
{"some string content"}
```

However, to account for situations where it may be desirable to embed `"}` within a string (usually, creating JSON), they also support heredoc-style delimiters, e.g.

```
{JSON"{"some_key": "some_value"}"JSON}
```

This modifies the lexer to tokenise these long strings as three different tokens, introducing `token.OPEN_LONG_STRING` and `token.CLOSE_LONG_STRING`. The parser handles these new tokens and the `ast.String` structure now captures the delimiter used so it's appropriately handled when emitting VCL in the formatter.

Previously, handling VCL using this long string syntax failed with an error message along the lines of:

```
Failed to run test: Parse Error: Undefined prefix expression for { at /tmp/example.vcl, line: 198, position: 15
```

It is now appropriately processed.

Thank you for your hard work on Falco, it is much appreciated!

### DIFF
--- a/ast/vcl_types.go
+++ b/ast/vcl_types.go
@@ -71,9 +71,9 @@ func (s *String) ID() uint64     { return s.Meta.ID }
 func (s *String) Expression()    {}
 func (s *String) GetMeta() *Meta { return s.Meta }
 func (s *String) String() string {
-	if s.Token.Offset == 4 { // offset=4 means bracket string
+	if s.LongString { // offset=4 means bracket string
 		return strings.TrimSpace(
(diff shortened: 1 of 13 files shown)


jcsaka (approved):

ysugimoto (approved):
@gabrielg Sorry for the late review, and I didn't know this spec.
I'll merge your changes because overall looks good and tidy up if I'm find something, thanks!

PR #380, opened 2024-10-29
Change `fastly.error` from `FLOAT` to `STRING`

gabrielg:
The Fastly docs indicate that [`fastly.error` is a `STRING`](https://www.fastly.com/documentation/reference/vcl/variables/miscellaneous/fastly-error/), but it's currently defined as a `FLOAT`. This changes the definition in `predefined.yml` to a string and regenerates the Go code.

Thank you!


ysugimoto (approved):
Oh I'm so rediculous... nice catch!

PR #384, opened 2024-11-15
Allows subroutine calls in return statments

gabrielg:
Not sure why this check was added, but Fastly, as of writing, _does_ allow subroutine calls in returns. For example, see these subroutines defined in `init`:

	sub return_test_str STRING {
		return std.toupper("test");
	}

	sub return_test_bool BOOL {
		return randombool(1, 2);
	}

These can be called from `vcl_recv` successfully like so:

	declare local var.strret STRING;
	declare local var.boolret BOOL;

	set var.strret = return_test_str();
	set var.boolret = return_test_bool();

	log "testing returns: " var.strret ", " var.boolret;

Here's an example Fiddle: https://fiddle.fastly.dev/fiddle/b872e5de

----

Has Fastly's behaviour perhaps changed recently and Falco is reflecting a previous behaviour?

### DIFF
--- a/interpreter/subroutine.go
+++ b/interpreter/subroutine.go
@@ -204,12 +204,6 @@ func (i *Interpreter) ProcessExpressionReturnStatement(stmt *ast.ReturnStatement
 	if err != nil {
 		return value.Null, NONE, errors.WithStack(err)
 	}
-	if !val.IsLiteral() {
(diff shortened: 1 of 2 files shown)


ysugimoto (approved):
I was not aware of this feature but Fastly Fiddle runs fine without error.
Fastly seems to improve this behavior as you pointed.

PR #385, opened 2024-11-22
Fix header subfield handling

gabrielg:
Multiple set calls for the same header subfield in a row, e.g.

	set req.http.Foo:bar = "baz";
	set req.http.Foo:bar = "snafu";

were resulting in multiple values for `bar` being present, e.g.

	"bar=baz, bar=snafu"

rather than the expected

	"bar=snafu"

You can confirm that this is Fastly's behaviour via this Fiddle:

https://fiddle.fastly.dev/fiddle/7af1b6c9


ysugimoto (approved):
Nice catch!

PR #389, opened 2024-12-10
update falco simulate command in README

TakeshiOnishi:
### Summary

This PR updates the falco command example in the README. The old command:
`falco local -debug /path/to/your/default.vcl`
is replaced with:
`falco simulate -debug /path/to/your/default.vcl`.

### Changes

- Updated the README file to replace `local` with `simulate` in the falco command example.

### Reason

The `simulate` command is now the correct and recommended subcommand for this context.  The `local` command may no longer be applicable or up-to-date.


ysugimoto (approved):
Thanks, you're right!!

PR #390, opened 2024-12-10
Add TIME type support for conditional operators to interpreter

Co9xs:
Supports TIME type in various Conditional operators.

Fastly VCL supports TIME type in Equal(=), Not equal(!=), Greater than(>), Less than(<), At least(>=), and At most(<=).

https://www.fastly.com/documentation/reference/vcl/operators/#conditional-operators


ysugimoto (approved):
Great Job!
Once I implementated this, it's not be able to compare `TIME` type... but it can do now, thanks!

PR #392, opened 2024-12-18
fix: support --version option

ronnnnn:
## Overview

1. Although the output of printHelp included a description of the --version flag[^1], it was not supported, so I added it.
2. Since the results of the -V and -h flags generally terminate with exit 0, I modified the exit status accordingly.

[^1]: https://github.com/ysugimoto/falco/blob/8ec8d6b9d3f0f4d68575da2d34ff2dc69c3583b8/cmd/falco/help.go#L65

### DIFF
--- a/cmd/falco/main.go
+++ b/cmd/falco/main.go
@@ -79,10 +79,10 @@ func main() {
 	}
 	if c.Help {
 		printHelp(c.Commands.At(0))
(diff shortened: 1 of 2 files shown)


ysugimoto (approved):
Thanks!
Yes the version and help command generally exit with `0`, you're right.

PR #439, opened 2025-05-01
feat(ast): support inline declarations in declare statements

acme:
Fastly now supports declarations with inline assignments: https://fiddle.fastly.dev/fiddle/0583d2c4

- Add optional value assignment to DeclareStatement
- Update parser to handle expressions after the type declaration
- Implement encoding/decoding support for the new value field
- Update string representation and formatting for declare statements with values
- Add comprehensive tests for the new functionality


ysugimoto (approved):
I'm excited to see this feature on Fastly!
Overall looks good, and additionally we need to implement this feature in the interpreter but I will do it on another PR, thanks.

ysugimoto:
Follows #444 for the interpreter support.

PR #440, opened 2025-05-01
feat(linter): add hit_for_pass to valid return values for vcl_fetch s…

acme:
…ubroutine

Although it's not very documented, Fastly supports `return(hit_for_pass)` in a `vcl_fetch()` subroutine. It is an alias for `return(pass)`.

It is mentioned briefly at:
https://docs.fastly.com/en/guides/understanding-the-different-pass-action-behaviors

Related: https://github.com/fastly/vscode-fastly-vcl/issues/26


ysugimoto (approved):
Looks good.
Also, this return the next state should be implemented on the interpreter, I will do it on the another PR.

PR #453, opened 2025-05-13
Function variable string conversion

ysugimoto:
Related https://github.com/ysugimoto/falco/pull/298

This PR support implicit string conversion on function calling.
Some builtin function accepts `STRING` type in argument, then the VCL could pass the other type (e.g INTEGR, BOOL) as variable indent like `var.IntValue`, `var.BoolValue`, and then Fastly converts the variable to `STRING` type implicitly.

The string type conversion is based on `(value.Value).String()` method and special case for a `NOTSET` string - treat as empty string.

### DIFF
--- a/__generator__/interpreter.go
+++ b/__generator__/interpreter.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"slices"
 	"sort"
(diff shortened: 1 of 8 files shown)


ysugimoto:
This fix will be released on the next major version.

PR #458, opened 2025-05-16
Add `-w` shorthand flag for Unit Testing watch mode

taga3s:
## description

The [falco docs](https://github.com/ysugimoto/falco/blob/main/docs/configuration.md) mentioned that Unit Testing watch mode could be enabled with both `--watch` and `-w` flags, but `-w` didn’t work. So, I added support for the flag, which I think could improve usability.


ysugimoto (approved):
Thanks, looks good.

PR #486, opened 2025-08-15
Fully implement and fix `fastly.try_select_shield()`

MasonM:
This completes the implementation of [fastly.try_select_shield()](https://www.fastly.com/documentation/reference/vcl/functions/miscellaneous/fastly-try-select-shield/) and fixes a minor bug with the fallback logic. I also added tests.


ysugimoto (approved):
LGTM

PR #489, opened 2025-08-23
Make querystring.get not decode query keys

yonran:
Make querystring.get not pct-decode the query keys as this does not match fastly behavior. But make it pct-decode and re-encode the key argument.

Pass through = in query parameter values instead of truncating values that have =.

Based on tests with fastly fiddle
https://fiddle.fastly.dev/fiddle/ed3feca5


ysugimoto (approved):
I was not aware of this behavior, nice catch!

PR #504, opened 2025-11-10
Add table.lookup_regex function + extensive tests

### LINKED ISSUE #503: [Feature] Support REGEX underlying type
### Kind of proposals

You can turn on multiple checkboxes.

- [x] Parser
- [x] Linter
- [x] Simulator
- [x] Testing

### Describe the feature

I have just found the additional `REGEX` underlying type in VCL which is documented here:
https://www.fastly.com/documentation/reference/vcl/types/regex/

We need to support this type for lexer, parser, linter and interpreter and need to implement `table.lookup_regex()` built-in function using this type.

jedisct1:
- Implements table.lookup_regex built-in for pattern-based table lookups
- Adds REGEX type support to type system and code generator
- Extends linter with regex validation helpers
- Enhances regsub/regsuball tests with additional test coverage

Fixes #503

### DIFF
--- a/interpreter/value/value.go
+++ b/interpreter/value/value.go
@@ -273,9 +273,9 @@ func (v *Acl) IsLiteral() bool { return v.Literal }
 func (v *Acl) Copy() Value     { return &Acl{Value: v.Value, Literal: v.Literal} }
 
 type Regex struct {
-	Value           string
-	Literal         bool
-	Unsatisfiable   bool // true if this is the unsatisfiable regex
+	Value         string
+	Literal       bool
+	Unsatisfiable bool // true if this is the unsatisfiable regex
 }
 
 func (v *Regex) String() string {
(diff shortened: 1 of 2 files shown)


interpreter/function/builtin/table_lookup_regex.go
```diff
@@ -0,0 +1,72 @@
...
+// Reference: https://developer.fastly.com/reference/vcl/functions/table/table-lookup-regex/
+func Table_lookup_regex(ctx *context.Context, args ...value.Value) (value.Value, error) {
+	// Argument validations
+	if err := Table_lookup_regex_Validate(args); err != nil {
+		return value.Null, err
+	}
+
+	id := value.Unwrap[*value.Ident](args[0]).Value
+	pattern := value.Unwrap[*value.String](args[1]).Value
+	defaultValue := value.Unwrap[*value.String](args[2])
```
ysugimoto: This function won't accpet a default value.

ysugimoto:
One of the important topic that we need discuss is about `unsatisfiable regex` variable.
This seems to be a never matched regex variable, and then,

1. How should we express the `unsatisfiable regex` on the `Type` on linter and `Value` on intepreter
2. On comparison, how should we evaluate it

Of couse we could define a new type of `REGEX` but I'm still under thinking.
We welcome your suggestion for treating the new type.

Issue #513, opened 2025-11-27
[Feature] add `assert.not_restart` test functionality for completeness


emylincon:
### Kind of proposals

You can turn on multiple checkboxes.

- [] Parser
- [] Linter
- [] Simulator
- [x] Testing

### Describe the feature

**required**
 Implement the `assert.not_restart()` function, which asserts the absence of a restart call (the opposite of `assert.restart()`). This is for completeness in testing.

Example usage:
```vcl
// @scope: recv
sub test_no_restart {
  testing.call_subroutine("some_subroutine");
  assert.not_restart();  // Passes if restart was NOT called
}
```

### Give us your considerations
Write some conditions, related links you've considered

PR #514, opened 2025-11-27
feat: add assert.not_restart function and tests 🆕

emylincon:
# Changes
- Implement assert.not_restart validation logic in assert_not_restart.go 🔧
- Add assert.not_restart documentation in testing.md 📚
- Create test cases for assert.not_restart in assert_not_restart_test.go 🧪
- Update assertion.test.vcl to include no_restart subroutine 🔄
- Register assert.not_restart in functions.go 📋

# Example

Assert restart statement has NOT been called.

```vcl
sub test_vcl {
    // vcl_recv will process normally without calling restart

    // Assert restart statement has not been called
    assert.not_restart();

    // With custom error message
    assert.not_restart("Expected no restart in this flow");
}
```

# Ticket
closes https://github.com/ysugimoto/falco/issues/513

### DIFF
--- a/tester/function/assert_not_restart.go
+++ b/tester/function/assert_not_restart.go
@@ -45,4 +45,3 @@ func Assert_not_restart(
 	}
 	return &value.Boolean{Value: true}, nil
 }
-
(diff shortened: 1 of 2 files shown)


ysugimoto (commented):
I agree with your thoughts, and implementations are also awesome!

Could you fix a tiny issue that is reported by `golangci-lint` plz?

PR #501, opened 2025-11-07
Support parameters in custom subs

jedisct1:
See https://www.fastly.com/blog/vcl-support-for-parameters-in-custom-subs

All the examples from the blog posts are passing.

Fixes #500

PR #516, opened 2025-12-04
formatter supports custom subroutine parameters

ysugimoto:
Related https://github.com/ysugimoto/falco/pull/501#issuecomment-3613224675

My apologies that I forgot to review for formatter implementation in #501, so we should support custom subroutine parameters on formatter.

I think we don't need additional formatter options for this, simply added formatting logic 🙃


ysugimoto:
I'll merge this PR for the patch version. Let me know if you nay any opinions

PR #521, opened 2025-12-18
feat: add support for linting VCL snippets with @scope annotation

acme:
Fastly VCL snippets are code fragments injected into specific subroutine phases rather than standalone VCL files with subroutine declarations. Previously, falco couldn't lint these files because they lack the `sub vcl_*` wrapper that provides scope context for variable validation.

The `@scope` annotation solves this by declaring which subroutine phase the snippet runs in, enabling the linter to validate that variables and functions are used correctly for that phase.

A snippet with a scope looks like:

```vcl
# @scope: deliver
unset resp.http.Server;
unset resp.http.X-Powered-By;
```

- Add `ParseVCLOrSnippet()` parser method that auto-detects whether input is full VCL or a snippet based on first token
- Add `IsSnippet` flag to `ast.VCL` to track parse mode
- Add `getFileLevelScope()` helper to extract `@scope` annotation from file-level comments
- Add `lintSnippetVCL()` to lint statements in the specified scope context
- Report error when snippet is missing required `@scope` annotation
- Update runner to use `ParseVCLOrSnippet()` for automatic detection

### DIFF
--- a/docs/linter.md
+++ b/docs/linter.md
@@ -124,7 +124,7 @@ falco lint snippet.vcl
 🔥 [ERROR] VCL snippet requires @scope annotation (e.g., # @scope: deliver) (snippet-scope-required)
 ```
 
-The available scope values are the same as for subroutine annotations: `recv`, `miss`, `hash`, `pass`, `fetch`, `error`, `deliver`, and `log`.
+The available scope values are the same as for subroutine annotations: `recv`, `hash`, `hit`, `miss`, `pass`, `fetch`, `error`, `deliver`, and `log`.
 
 ## Linter Plugin
 


parser/parser.go
```diff
@@ -210,6 +210,44 @@ func (p *Parser) peekPrecedence() int {
...
+
+	// Parse as snippet - the file contains statements, not declarations
+	statements, err := p.ParseSnippetVCL()
+	if err != nil {
+		return nil, err
+	}
+
+	// Return VCL with snippet flag set
+	vcl := &ast.VCL{
+		Statements: make([]ast.Statement, len(statements)),
```
jedisct1: The slice isn't reused, to that could simply be `Statements: statements`.

PR #525, opened 2025-12-27
Req hash handling fix

akrainiouk:
This PR reimplements '+=' operator for _req.hash_ variable relaxing type requirements so that the behavior better matches fastrly.
Allows for STRING and BOOL literals to be used as arguments as well as variables of all types except for

### DIFF
--- a/interpreter/interpreter.go
+++ b/interpreter/interpreter.go
@@ -580,23 +580,6 @@ func (i *Interpreter) ProcessFetch() error {
 	}
 	// TODO: consider stale-white-revalidate and stale-if-error TTL
 
-	// Update cache
-	defer func() {
-		resp := i.ctx.BackendResponse.Clone()
-		// Note: compare BackendResponseCacheable value
-		// because this value will be changed by user in vcl_fetch directive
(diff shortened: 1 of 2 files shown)


interpreter/assign/addition.go
```diff
@@ -9,6 +11,21 @@ import (
 	"github.com/ysugimoto/falco/interpreter/value"
 )
 
+func UpdateHash(left *value.String, right value.Value) error {
+	if right.Type() != value.StringType && right.Type() != value.BooleanType && right.IsLiteral() {
+		return errors.WithStack(fmt.Errorf("only STRING and BOOL literals are allowed, got %s", right.Type()))
```
jedisct1: The test reads "reject if (not STRING and not BOOL and is literal)", so integer variables are allowed . Is that intentional? That doesn't match the error message.

jedisct1:
> Makes sense (still didn't get an answer from fastly folks BTW)

If the hash is used as a cache key, it’s very likely that the input includes things like a secret key and the service and customer IDs, ensuring proper context separation. So even knowing the algorithm wouldn’t be useful for Falco.

PR #526, opened 2025-12-27
Support for headers property in linter

akrainiouk:
Turns out `headers` property of `req`, `resp`, `beresp` and `obj` variables is not hooked up in falco linter.
This PR adds necessary definitions.


ysugimoto (changes requested):
You could not change the `linter/context/predefined.go` because this file is auto-generated via generator script.
Please add `xxx.headers` to yaml file on https://github.com/ysugimoto/falco/blob/main/__generator__/predefined.yml and generate them.

akrainiouk:
Changed as suggested.
Thanks for the pointer.

Issue #523, opened 2025-12-26
[BUG] Multiple conditional expressions with && operator do not work correctly


bungoume:
### Kind of proposals
- [] Parser
- [] Linter
- [] Simulator
- [x] Testing

### Describe the problem
When combining multiple conditions using the && operator in an if statement, the expression evaluates incorrectly and returns false, even though each individual condition evaluates to true when tested separately.

Specifically:

* An empty string variable evaluates to true ✅
* Negation of an undefined header (!req.http.UNDEF) evaluates to true ✅
* However, combining these with && incorrectly evaluates to false ❌

### VCL code that cause the problem / reproduceable
```
sub test_empty_recv {
  declare local var.EMPTY STRING;
  set var.EMPTY = "";

  set req.http.EXPECT-TRUE-1 = if (var.EMPTY, "1", "0");
  set req.http.EXPECT-TRUE-2 = if (!req.http.UNDEF, "1", "0");
  set req.http.EXPECT-TRUE-3 = if (var.EMPTY && !req.http.UNDEF, "1", "0");

  assert.equal(req.http.EXPECT-TRUE-1, "1");  // PASS
  assert.equal(req.http.EXPECT-TRUE-2, "1");  // PASS
  assert.equal(req.http.EXPECT-TRUE-3, "1");  // FAIL
}
```

### Expected behavior
All three assertions should pass
https://fiddle.fastly.dev/fiddle/a721c914

### Desktop (please complete the following information)

 - OS: Mac
 - Shell: zsh

### Additional context
This issue can be reproduced in either of the following versions:

- Falco v2.0.0
- Falco v1.21.1

PR #527, opened 2025-12-27
Fix logical condition with notset string

ysugimoto:
fixes: https://github.com/ysugimoto/falco/issues/523

Since I implemented `notset` string, I forgot to pay attention for logical operator expression evaluation.
Before, we simply compared with empty string like `left.value != ""` but this implementation causes unexpected behavior. This PR fixes this problem and completely support. 

On comparing left and right expressions with using logical operator (`&&` or `||`), we explicitly check whether the string value is notset, especitally singular expression (e.g request headers or HTTP request/response header value)


interpreter/operator/operator_logical_or_test.go
```diff
@@ -122,6 +126,8 @@ func TestLogicalOrOperator(t *testing.T) {
 			{left: &value.String{Value: "example"}, right: &value.String{Value: "example", Literal: true}, isError: true},
 			{left: &value.String{Value: "example"}, right: &value.String{Value: ""}, expect: true},
 			{left: &value.String{Value: "example"}, right: &value.String{Value: "", Literal: true}, isError: true},
+			{left: &value.String{Value: "example"}, right: &value.String{Value: ""}, expect: true},
```
jedisct1: This test is the same as line 127.

jedisct1:
Some tests are duplicated, but otherwise LGTM!

PR #530, opened 2026-01-09
Stringify backends and boolean literals

akrainiouk:
stringifyVariableArguments rejects arguments of type BACKEND as well as boolean literals (true/false) while
fastly does support both.
Backend is converted to corresponding VCL ID string
true/false are converted to "1"/"0" respectively
Fastly behavior is demonstrated in this fiddle: https://fiddle.fastly.dev/fiddle/333f8d33

This PR brings Falco behavior in compliance with Fastly

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -688,6 +688,7 @@ Set fixed access rate. This function affects to `ratelimit` related values and f
 - `ratecounter.rc.rate.60s`
 - `ratelimit.check_rate()`
 - `ratelimit.check_rates()`
+- `ratelimit.ratecounter_increment()`
 
 ```vcl
 ratecounter rc {}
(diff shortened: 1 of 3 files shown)


interpreter/function/builtin_functions.go
```diff
@@ -34,15 +34,12 @@ func stringifyVariableArguments(name string, args []value.Value, indicies map[in
 			} else {
 				stringified[i] = &value.String{Value: t.Value.String()}
 			}
-		case *value.Backend:
-			// Backend value could not convert to string
-			return nil, fe.CannotConvertToString(name, i)
```
akrainiouk: Backend value is treated the same as other compatible types

ysugimoto (approved):
Looks good.

PR #543, opened 2026-01-22
Fix random_* functions

jedisct1:
- randomstr: fix off-by-one error where last character was never selected and fix default charset typo (missing g/G, duplicate d/D)
- randomstr: return `null `for empty charset
- randombool: use integer-based comparison instead of float
- randombool: return true when `denom <= 0` and `num > 0`
- randomint: return 0 when `from > to`

Also, seeding is not necessary any more in current Go versions, and this is not a good idea before every random call.


ysugimoto (commented):
Thanks for the improvements! and I put some commens.

I ensured these fixes match the Faslty behavior: https://fiddle.fastly.dev/fiddle/fd7218f1

PR #580, opened 2026-03-12
fix(simulate): prevent response headers from appearing as body in proxy mode

3846masa:
## Summary

Fix `falco simulate --proxy` returning HTTP response headers as body content, a regression introduced by #509.

## Problem

#509 changed `sendResponse()` to use `(*http.Response).Write(w)` to preserve custom status text set via `obj.response`. However, `Response.Write()` writes the full raw HTTP response (status line + headers + body) into the `io.Writer`. Since Go's `http.ResponseWriter` auto-generates its own status line and headers on the first `Write()` call, the raw output ended up appearing as the response body.

## Fix

Split response writing into two paths:

- **HTTP/1.x** (`sendRawResponse`): Hijack the TCP connection and write the raw HTTP response directly, preserving custom status text from `obj.response`.
- **HTTP/2** (`sendFormattedResponse`): Use the standard `http.ResponseWriter` API to write headers and body separately. Custom status text is not preserved, but this is correct because HTTP/2 has no reason phrase in the protocol.

## What is preserved from #509

- `synthetic` and `obj.response` independence (removed `HasSyntheticResponse` flag)
- Default Fastly headers set before `vcl_deliver` (allowing them to be unset by VCL)
- Custom status text via `obj.response` (on HTTP/1.x connections)

---

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/interpreter/handler.go
+++ b/interpreter/handler.go
@@ -89,15 +89,45 @@ func (i *Interpreter) sendResponse(w ghttp.ResponseWriter) {
 		return
 	}
 
-	// We have to use (*http.Response).Write(io.Writer)
-	// in order to write custom status text which is set via obj.response variable.
-	//
-	// Typically, on handling HTTP, http.ResponseWriter interface is enough to respond regular HTTP specs
-	// but could not set custom status text on the status line.
-	//
-	// Fortunately (*http.Response).Write(io.Writer) method will support to output custom status text
-	// so we need to send HTTP response via this method.
-	i.ctx.Response.Write(w) // nolint:errcheck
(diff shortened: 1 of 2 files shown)


ysugimoto (commented):
This is good to know for me, thanks.
Then could you add some test to ensure the custom status test is preserved?

ysugimoto (approved):
Great, thanks!

ysugimoto:
@3846masa This PR is ready to merge so could you merge the upstream branch to resolve out-of-date?

PR #581, opened 2026-03-15
fix: Fix GetField to correctly handle quoted strings with escaped quotes

3846masa:
## Summary

- Fix `GetField` to correctly handle quoted strings containing escaped quotes (e.g. `"\"asdf\""`)
- Previously, `strings.Trim(val, "\"")` would strip all leading/trailing `"` characters greedily, breaking values like `"\"asdf\""` which should resolve to `"asdf"`
- Now only the outermost quotes are removed, and inner escaped quotes (`\"`) are unescaped properly

## Changes

- field.go
  - Replace `strings.Trim` with explicit prefix/suffix check and `strings.ReplaceAll` for escaped quote handling
- field_test.go
  - Add test case for quoted string with escaped quotes

## Test plan

- [x] Existing `TestGetField` cases continue to pass
- [x] New test case `b="\"asdf\""` correctly returns `"asdf"`

### DIFF
--- a/interpreter/variable/field.go
+++ b/interpreter/variable/field.go
@@ -36,8 +36,8 @@ func GetField(subject, key, sep string) *value.String {
 
 	// Due to fastly behavior with header values that do not conform to RFC-8941
 	// we don't always want to strip quotes.
-	if strings.HasPrefix(val, "\"") {
-		val = strings.Trim(val, "\"")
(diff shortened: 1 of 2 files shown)


ysugimoto (commented):
Looks good, but I would like to know the actual Fastly behavior.
Could you put the Fastly behavior on the fiddle?

ysugimoto (approved):

ysugimoto:
@3846masa This PR is ready to merge so could you merge the upstream branch to resolve out-of-date?

PR #584, opened 2026-03-22
fix: normalize all TIME values to UTC

bungoume:
## Summary
- Fastly handles TIME values in UTC (GMT), but several places used `time.Unix()` or `time.Now()` without `.UTC()`, causing local timezone-dependent values in `value.Time`
- Added `.UTC()` to all `time.Time` construction and assignment points to ensure consistent environment-independent behavior
- Affected areas: builtin functions (`std.time`, `std.integer2time`, `time.hex_to_time`, `time.add`, `time.sub`), variable getters (`now`, `time.start`, `time.end`, etc.), assignment operations (`TIME = INTEGER/FLOAT/RTIME`), context initialization, and testing fixed time


acme:
Good point, UTC all the things! However, I think there are a few more changes necessary in operator.go, interpreter.go, and the Date header.

acme (approved):
Thanks! I implemented it in a slightly different way.

PR #585, opened 2026-03-24
feat(formatter): add CLI flags for all format options

acme:
- Add `cli` struct tags to all 16 FormatConfig fields so they are accessible as `--flag` arguments to `falco fmt`
- Add missing `breakCompoundConditions` and `trailingCommentWidth` to Wasm format API (defaultFormatConfig + applyFormatOptions)
- Add break compound conditions checkbox to Wasm HTML demo
- Update help text, formatter docs, and configuration docs


jedisct1:
This works, but I wonder if 16 CLI flags (and growing) is the right UX here.

`.falco.yml` already supports all these options. and that's a better fit for settings that are usually project-level and don't change between runs. 

Instead of adding flags for every option, what about just documenting the config file approach better and maybe adding a `falco fmt --init` command that generates a `.falco.yml` file with the format defaults?

acme:
I still think it's useful having multiple ways to set a formatting option. For example:

- I want to try the different options, without having to edit a file.
- I want to call this in my VS Code Editor plugin with a slightly different default, without creating a config file.
- I want to call this in Wasm, which doesn't have a configuration file.

acme:
Alternatively we could make breakCompoundConditions the default, which is all that I'm really using this for.

jedisct1:
If `breakCompoundConditions` is the main use case, the simplest path would be to just change its default to `true`. What do you think @ysugimoto ?

ysugimoto:
@jedisct1 I agree. Setting it to `true` by default would probably improve readability.
Let's announce this change on the next version release note 👍

acme:
Great, I've opened https://github.com/ysugimoto/falco/pull/593 and will close this.

PR #586, opened 2026-03-24
Handling null in assignment to header field

akrainiouk:
Brings to compiance with Fastly various cases of (null) handling in set statement.

Namely:
* assignment to http header field is treated the same way as an assignment to local variable (i.e. null is interpreted as empty string)
* reassignment of null to an existing header value is not ignored but instead leads to resetting that header to null

For reference, Fastly behavior is reproduced in this fiddle: https://fiddle.fastly.dev/fiddle/be239b5c


interpreter/variable/header.go
```diff
@@ -62,8 +62,10 @@ func getResponseHeaderValue(r *http.Response, name string) *value.String {
 func setRequestHeaderValue(r *http.Request, name string, val value.Value) {
 	name, key, found := strings.Cut(name, ":")
 	if !found {
-		// Skip when set value is notset
+		// Setting header to undefined value is equivalent to deleting it
 		if isNotSetValue(val) {
+			r.Header.Del(name)
+			r.Unassign(name)
```
akrainiouk, line 68: The intent is to delete HTTP header, but I am not entirely confident this is
the correct way to do it.

jedisct1, line 68: `setResponseHeaderValue` has the same issue, so it needs the same fix.

interpreter/statement.go
```diff
@@ -266,6 +268,28 @@ func (i *Interpreter) ProcessSetStatement(stmt *ast.SetStatement) error {
...
+func (i *Interpreter) ProcessSetStatementHeaderField(stmt *ast.SetStatement) error {
+	left, err := i.vars.Get(i.ctx.Scope, stmt.Ident.Value)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	if err := isValidStatementExpression(left.Type(), stmt.Value); err != nil {
+		return errors.WithStack(err)
+	}
+	// for header fields, undefined values are treated the same as for local variables,
+	// thus we interpret left hand side as local variable expression
```
jedisct1, line 280: Do you mean right hand side?

interpreter/set_statement_test.go
```diff
@@ -0,0 +1,225 @@
...
+		v, err := ip.vars.Get(context.RecvScope, "req.http.x-header")
+		if err != nil {
+			t.Errorf("var.foo must be declared: %s", err)
+			return
+		}
+		assertValuesEqual(t, &value.String{IsNotSet: true}, v)
+		assertStringsEqual(t, "(null)", ip.process.Logs[0].Message)
+	})
+
+	t.Run("json.excape of null to local var", func(t *testing.T) {
```
jedisct1, line 110: Typo :)

jedisct1 (changes requested):

interpreter/variable/header.go
akrainiouk: Thanks for spotting. Fixed.

interpreter/set_statement_test.go
akrainiouk: Fixed, here and in one more place

interpreter/statement.go
akrainiouk: I actually did mean left hand side but the wording was awkward and confusing.
Hopefully fixed version is clearer.

interpreter/statement.go
```diff
@@ -32,8 +32,7 @@ func nopSeekCloser(r io.ReadSeeker) io.ReadSeekCloser {
 func (i *Interpreter) ProcessBlockStatement(
 	statements []ast.Statement,
 	ds DebugState,
-	// isReturnAsValue indicates return statement may return value.Value in functional subroutine if true
-	isReturnAsValue bool,
+	isReturnAsValue bool, // if true indicates return statement may return value.Value in functional subroutine
```
akrainiouk, line 35: had to modified the comment to make linter happy

akrainiouk:
@jedisct1 thanks for looking! Your comments are now addressed.

acme (approved):
Thanks! I implemented it a little differently.

PR #590, opened 2026-03-31
if_statement: fix else-if comment restored from the wrong variable

jedisct1:
The loop over `else-if` branches was restoring from the outer 'tmp' variable (which holds the main if-block's comments),instead of the value saved from the current else-if branch.


acme:
Could you please add a test for this?

jedisct1:
Added.

acme (approved):
👍

PR #591, opened 2026-04-01
fix: remove incorrect table/item-limitation lint

acme:
- The 1000 item limit applies to Edge Dictionaries, not inline VCL tables
- Remove TABLE_ITEM_LIMITATION constant and reference URL
- Remove the check from lintTableDeclaration
- Remove documentation from rules.md


jedisct1:
The Fastly documentation at https://www.fastly.com/documentation/reference/vcl/declarations/table/#limitations mentions a 1000-item default limit for tables. Is that documentation outdated?

If the limit is real but just configurable via Fastly support, we could make the threshold configurable instead of removing the check. What do you think?

acme:
The limit applies to edge dictionaries, not tables defined in VCL. I have raised a Fastly internal ticket to clarify the documentation.

acme:
Fastly updated https://www.fastly.com/documentation/reference/vcl/declarations/table/#limitations to remove the 1,000 limit text and instead say:
> All tables combined support up to 100,000 items.

jedisct1 (approved):

PR #596, opened 2026-05-06
fix(interpreter): return NOT SET from std.strstr on miss

acme:
- Return `&value.String{IsNotSet: true}` instead of an empty string when the needle is not found, matching Fastly's documented behavior https://www.fastly.com/documentation/reference/vcl/functions/strings/std-strstr/index.md
- Fixes `if (std.strstr(...))` being unconditionally truthy: an empty-but-set string is truthy in VCL boolean context, only NOT SET is falsy
- Add `TestIfStatement` regression case covering the reported `if (std.strstr(req.url, "/nothing-that-matches"))` scenario

### DIFF
--- a/interpreter/function/builtin/std_strstr.go
+++ b/interpreter/function/builtin/std_strstr.go
@@ -41,7 +41,10 @@ func Std_strstr(ctx *context.Context, args ...value.Value) (value.Value, error) 
 
 	idx := strings.Index(haystack, needle)
 	if idx == -1 {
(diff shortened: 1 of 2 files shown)


akrainiouk (approved):
LGTM (Verified with fastly fiddle):
```varnish
sub vcl_recv {
  if (std.strstr("foo", "bar")) {
    log "result is set";
  } else {
    log "result is unset";
  }
}
```

PR #598, opened 2026-05-07
fix(interpreter) null to string conversion in sub argument evaluation

akrainiouk:
In the context of sub (function and procedure) argument evaluation null values should be treated the same way
as in set statement context i.e. converted to blank string rather than '(null)' (current state)
See this fastly fiddle: https://fiddle.fastly.dev/fiddle/10f2511b

The fix explicitly passes `LocalVariableExpression()` option to `i.ProcessExpression` function call when processing function and procedure arguments.
In the log context null is treated as before i.e. converted to '(null)' string.


interpreter/expression_test.go
```diff
@@ -352,6 +352,38 @@ func TestProcessExpression(t *testing.T) {
...
+            sub func(STRING var.value) STRING {
+                set req.http.x-func-arg = var.value;
+                return var.value;
+            }
+            sub vcl_recv {
+                declare local var.null STRING;
+                declare local var.result STRING = func("[" var.null "]");
+			}`,
+			assertions: map[string]value.Value{
+				"req.http.x-func-arg": &value.String{Value: "[]"},
```
akrainiouk, line 374: used to evaluate to  [(null)]

interpreter/statement_test.go
```diff
@@ -476,6 +476,21 @@ func TestFunctionCallStatement(t *testing.T) {
...
+			vcl: `
+                sub proc(STRING var.value) {
+                    set req.http.x-proc-arg = var.value;
+                }
+				sub vcl_recv {
+					declare local var.null STRING;
+					call proc("[" var.null "]");
+				}`,
+			assertions: map[string]value.Value{
+				"req.http.x-proc-arg": &value.String{Value: "[]"},
```
akrainiouk, line 490: used to evaluate to  [(null)]

interpreter/statement.go
```diff
@@ -352,7 +351,7 @@ func (i *Interpreter) ProcessCallStatement(stmt *ast.CallStatement, ds DebugStat
 	// Evaluate arguments
 	var args []value.Value
 	for _, argExpr := range stmt.Arguments {
-		argVal, err := i.ProcessExpression(argExpr)
+		argVal, err := i.ProcessExpression(argExpr, LocalVariableExpression())
```
akrainiouk, line 362: treating call statement the same as variable assignment (set statement)

interpreter/statement.go
```diff
@@ -32,8 +32,7 @@ func nopSeekCloser(r io.ReadSeeker) io.ReadSeekCloser {
 func (i *Interpreter) ProcessBlockStatement(
 	statements []ast.Statement,
 	ds DebugState,
-	// isReturnAsValue indicates return statement may return value.Value in functional subroutine if true
-	isReturnAsValue bool,
+	isReturnAsValue bool, // if true indicates return statement may return value.Value in functional subroutine
```
akrainiouk, line 35: this is to make linter happy

interpreter/expression.go
```diff
@@ -291,7 +291,7 @@ func (i *Interpreter) ProcessFunctionCallExpression(exp *ast.FunctionCallExpress
 				)
 			}
 		} else {
-			a, err := i.processExpression(exp.Arguments[j], opt)
+			a, err := i.ProcessExpression(exp.Arguments[j], LocalVariableExpression())
```
akrainiouk: treating function invocation context the same way as variable assignment (set statement)

acme (approved):
Thanks, looks good.

I made a little change and verified it matches Fastly.

PR #600, opened 2026-05-13
feat(linter): add fastly_info.version predefined variable

togdon:
Adds fastly_info.version (STRING, read-only, all scopes), which records the currently deployed VCL version number and is commonly used in logging expressions.

Regenerates linter/context/predefined.go, interpreter/variable/predefined.go, and console/suggestions.go from __generator__/predefined.yml.


togdon:
@acme we've had this variable in our internal Fastly code since at least 2018 (exporting it out to Sumo Logic, which at one point documented its presence [here](https://www.sumologic.com/help/docs/integrations/saas-cloud/fastly/)). In trying to get linting working better for us we've found that it's missing in Falco, but **_also_** looks like it's missing in the Fastly Docs as well at this point. 

It may be that we've got some old lingering code hanging around and this PR should be closed (and we should drop `fastly_info.version` from our logs and just focus on `req.vcl.version` (which we also log/use)), we do, however, still see it populated today (looks to be a SHA-1 hash that is consistent across POPs for similar request types). I'm guessing you may be one of the few folks looking at this repo who can somewhat definitively answer if it's still something that exists and is worth being included in Falco

acme:
The STRING `fastly_info.version` variable is intentionally internal to Fastly and customers shouldn't depend on it.

acme (approved):
We do support a number of other undocumented variables, so let's merge it in. Thanks!

PR #601, opened 2026-05-21
Added location properties in tester json output

akrainiouk:
Added File, Line and Position properties to tester's JSON output, so that it is possible to properly locate failure.


acme (changes requested):
Thanks, would you be able to add a test?

akrainiouk:
> Thanks, would you be able to add a test?

Added unit tests for TestCase JSON serialization: 
https://github.com/akrainiouk/falco/blob/19aebf2ee3979630d35886e660265e094874b68f/tester/entity_test.go

Also to make it more consistent with other properties, I modified added property types changing them from pointers to actual values.

acme (approved):
Thanks, looks great.

I improved the tests a little.

PR #604, opened 2026-05-27
function parameter and return value handling cleanup

akrainiouk:
This PR brings function argument and return type handling in better consistency with Fastly. It also fixes several issues with handling arguments of `REGEX` type.

1. Original code was overly aggressive in avoiding argument copying, as a result literal arguments were retaining their status even inside function body which lead to unexpected behavior. Specifically literal arguments were treated as literal inside subroutine body while in reality they should have been turned into variables.
2. There were certain type combinations that were rejected in Falco even though there are allowed in Fastly

This PR modifies `convertValueToType` function so it relies on `assign.Assign` function for implicit conversions only enforcing additional restrictions that are specific to function arguments and return values.
This should bring conversion behavior into consistency between set statement and function invocation implementation.

This fastly fiddle was used to investigate the details of argument handling: https://fiddle.fastly.dev/fiddle/c3494cd8
Commented out code represents those additional limitations imposed by function arguments as compared to variable assignment.


interpreter/assign/assign.go
```diff
@@ -247,6 +252,27 @@ func Assign(left, right value.Value) error {
...
 		}
+	case value.RegexType:
+		lv := value.Unwrap[*value.Regex](left)
+		switch right.Type() {
+		case value.StringType: // REGEX = STRING
+			rv := value.Unwrap[*value.String](right)
+			if !rv.IsLiteral() {
+				return errors.WithStack(fmt.Errorf("string value must be a literal for REGEX assignment"))
+			}
+			_, err := pcre.Compile(rv.Value)
```
akrainiouk, line 263: literal strings are pre-validated when coverted to REGEX upon assignment.

interpreter/assign/assign_test.go
```diff
@@ -318,4 +318,38 @@ func TestProcessAssignment(t *testing.T) {
 			}
 		}
 	})
+
+	t.Run("left is REGEX", func(t *testing.T) {
```
akrainiouk, line 322: test for newly implemented automatic conversions to REGEX

interpreter/value/value.go
```diff
@@ -57,6 +58,36 @@ type Value interface {
 
 type null struct{}
 
+// Create creates new value of the specified type.
+func Create(valueType Type) (Value, error) {
```
akrainiouk, line 62: This function is factored out of local.Declare function as it is now also used for creation of function arguments and return values upon type conversion.

interpreter/interpreter_test.go
```diff
@@ -69,7 +69,7 @@ func assertInterpreter(t *testing.T, vcl string, scope context.Scope, assertions
 			return
 		}
 		if diff := cmp.Diff(val, v); diff != "" {
-			t.Errorf("Value assertion error, diff: %s", diff)
+			t.Errorf("Value assertion error for '%s', diff: %s", name, diff)
```
akrainiouk, line 72: Reporting improvement, so that upon failure corresponding VCL variable is also printed along with diffs.

interpreter/subroutine.go
```diff
@@ -206,20 +205,17 @@ func (i *Interpreter) ProcessFunctionSubroutine(sub *ast.SubroutineDeclaration,
...
-				}
-				val = converted
+			converted, err := convertValueToType(val, expectedType)
+			if err != nil {
+				return val, NONE, exception.Runtime(
+					&t.GetMeta().Token,
+					"Invalid return type, expects=%s, but got=%s. Conversion failed: %s",
+					sub.ReturnType.Value,
+					val.Type(),
+					err.Error(),
```
akrainiouk, line 215: Original code was ignoring the underlying error which some times lead to missing information when conversion failed and confusing error messages.

interpreter/subroutine.go
```diff
@@ -341,17 +337,15 @@ func (i *Interpreter) validateAndSetParameters(sub *ast.SubroutineDeclaration, a
...
+		converted, err := convertValueToType(arg, value.Type(param.Type.Value))
 		if err != nil {
 			return exception.Runtime(
 				&param.GetMeta().Token,
-				"Parameter %s expects type %s, but got %s",
+				"Invalid parameter %s, expects=%s, but got=%s. Conversion failed: %s",
 				param.Name.Value,
 				param.Type.Value,
 				arg.Type(),
+				err.Error(),
```
akrainiouk, line 348: Same as above. Original error included in the message.

acme (approved):
Thanks, looks good.

I made a few small changes.

PR #606, opened 2026-06-07
feat: Extend `testing.call_subroutine` With Arguments and Return Values

### LINKED ISSUE #602: [Feature]: Extend `testing.call_subroutine()` with Parameter Support
### Kind of proposals

- [x] Linter
- [x] Testing
- [ ] Parser
- [ ] Simulator

### Describe the feature

Currently, `testing.call_subroutine()` only accepts the subroutine name as an argument and cannot pass parameters to the target subroutine. This limits test flexibility when testing functional subroutines with parameters or scoped subroutines that require arguments.

Proposed functionality:
1. **Passing arguments to subroutines under test** — functional and scoped subroutines with declared parameters can receive values via `testing.call_subroutine()`
2. **Typed return values for functional subroutines** — `testing.call_subroutine()` returns the actual declared return type (BOOL, STRING, etc.) instead of always returning a state string
3. **Type validation at lint time** — the linter validates extra arguments against the target subroutine's parameter types before runtime

**Example VCL After Implementation**

```vcl
// Declare a functional subroutine that accepts parameters
sub url_path_is_either(STRING var.path1, STRING var.path2) BOOL {
  if (req.url.path == var.path1 || req.url.path == var.path2) {
    return true;
  }
  return false;
}


// Test file — now can pass arguments and capture return values
// @scope: recv
sub test_path_matches {
  set req.url = "/path1/foo";
  declare local var.result BOOL;
  
  // Pass arguments to the subroutine and capture the BOOL return value
  set var.result = testing.call_subroutine("url_path_is_either", "/path1/foo", "/path2/bar");
  assert.true(var.result);
}
```

**Related Infrastructure**

- The interpreter already supports subroutine arguments via `ProcessSubroutine()` and `ProcessFunctionSubroutine()` accepting `args []value.Value`
- The interpreter already performs implicit type conversions and validates parameter types via `validateAndSetParameters()`
- The parser already supports functional subroutine declarations with return types: `sub name(TYPE var, ...) RETURN_TYPE { ... }`

emylincon:
## Overview

This PR extends `testing.call_subroutine` to support passing typed arguments to subroutines that declare parameters, and capturing the return value of functional subroutines.

Previously, `testing.call_subroutine` accepted exactly one argument and  always passed `nil` args to the interpreter,
making it impossible to exercise functional subroutines with declared parameters from within a test. The existing parser and interpreter already support subroutine parameter and return-type syntax; this work connects the testing layer to the existing code.

Backward compatibility is fully preserved — all existing single-argument calls remain valid without any changes.

### Documentation

Updated`testing.call_subroutine` in `docs/testing.md` with new syntax

## Implementation Phases

| # | Phase | Summary |
|---|-------|---------|
| 1 | Signature update | Remove static arg-type slice; make validator variadic; slice `args[1:]` as `subArgs` |
| 2 | Return value propagation | Introduce `CallResult`; capture and return functional subroutine values |
| 3 | Caller-side wiring | Update `functions.go` wrapper to branch on `IsFunctional`; forward `subArgs` to interpreter |
| 4 | Linter awareness | Relax argument-count rule; add `Extra` hook for dynamic parameter validation |
| 5 | Formatter verification | Confirm no structural formatter changes needed; add multi-argument formatting test case |

## Example Usage

```vcl
// Functional subroutine under test
sub url_path_is_either(
    STRING var.path1, STRING var.path2
) BOOL {
  if (req.url.path == var.path1
      || req.url.path == var.path2) {
    return true;
  }
  return false;
}

// Test exercising the new syntax
declare local var.result BOOL;
set var.result = testing.call_subroutine(
    "url_path_is_either", "/path1", "/path2"
);
assert.true(var.result);

// Scoped subroutine call (unchanged behaviour)
assert.state(lookup);
```

### Return value behaviour

| Subroutine kind | Return value |
|:----------------|:-------------|
| Scoped (lifecycle) | `STRING` state name (e.g. `"lookup"`) |
| Functional (has `ReturnType`) | The typed value returned by the subroutine |

# Ticket
closes https://github.com/ysugimoto/falco/issues/602


acme (approved):
Thanks, looks good.

I made a few changes.

PR #608, opened 2026-06-11
Implemented url.normalize built-in function

akrainiouk:
See: https://www.fastly.com/documentation/reference/vcl/functions/strings/url-normalize/


acme (approved):
Thanks, looks great!

PR #612, opened 2026-06-16
fix: return req/beresp.backend.port as INTEGER

mteodoro:
## Problem

  - `req.backend.port` (deliver/error/log) parsed the backend `.port` property
    via `strconv.ParseInt(p.Value.String())`, but `String()` returns the quoted
    VCL form (`"443"`), which `ParseInt` rejects — yielding `value.Null` and a
    runtime exception on any subsequent use.
  - `beresp.backend.port` (fetch) returned a `*value.String`, contradicting its
    declared INTEGER type, and lacked a nil-backend guard.

  ## Fix

  Add a `getBackendPort` helper in `shared.go` that reads the `*ast.String.Value`
  directly and returns `*value.Integer`; all four scopes call it. Adds a unit
  test covering each scope.

  Verified against Fastly docs:
  - https://www.fastly.com/documentation/reference/vcl/variables/miscellaneous/req-backend-port/
  - https://www.fastly.com/documentation/reference/vcl/variables/backend-connection/beresp-backend-port/

### DIFF
--- a/interpreter/variable/backend_test.go
+++ b/interpreter/variable/backend_test.go
@@ -1,0 +1,80 @@
+package variable
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/ysugimoto/falco/ast"
+	"github.com/ysugimoto/falco/interpreter/context"
+	"github.com/ysugimoto/falco/interpreter/value"
+)
+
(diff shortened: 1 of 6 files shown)


acme (approved):
Thanks, looks good.

I verified the behaviour with Fiddle and added a test case.

PR #613, opened 2026-06-17
Make += append to HTTP headers instead of overwriting them

jedisct1:
Header assignment ignored the operator in every scope.

`set req.http.X += "y"` replaced the value rather than appending to it.

Fix it.


akrainiouk (approved):
Looks good. Thanks!

PR #614, opened 2026-06-17
Enforce the VCL call tree and request workspace limits

jedisct1:
Fastly rejects VCL whose fully inlined call graph exceeds 25000 calls, and returns a 503 once the 256 KB request workspace is exhausted.

Do the same thing to avoid scripts that work in Falco, but not on Fastly.

### DIFF
--- a/interpreter/context/context.go
+++ b/interpreter/context/context.go
@@ -79,6 +79,11 @@ type Context struct {
 	RequestStartTime time.Time
 	CacheHitItem     *cache.CacheItem
 
+	// RequestWorkspaceBytes tracks how much of the per-request workspace has been
+	// consumed by assembling request headers. Fastly never reclaims it within a
+	// request, not even across restarts, so this is never reset once set.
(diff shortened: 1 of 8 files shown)


acme (changes requested):
Thanks, this was on my TODO list.

This charges value + name + 9 per request header write, but in my Fiddle tests Fastly charges value + name + 3 (without the `req.http.`). Could you match this?

In a Fiddle, Fastly consumes ~14 KB of workspace before any user VCL executes. I'm not sure if we should match this, as Fiddle includes observation overhead.

jedisct1:
Thanks!

The computation should now be more accurate.

In real fastly services, the overhead is about 8.5 KB. So, to be conservative we can use 10 KB. All the inbound headers are also counted.

acme (approved):
Thanks, looks great.

acme:
I can't seem to update branch or merge. Can you?

jedisct1:
Yep. I'll do it.

PR #593, opened 2026-04-24
feat: default break_compound_conditions to true

acme:
- Change default from false to true in FormatConfig
- Add BreakCompoundConditions to wasm FormatOptions
- Wire up JS option mapping in applyFormatOptions
- Update config test to expect new default

PR #615, opened 2026-06-18
fix: add /v2 module path suffix

acme:
The module is tagged v2.0.0+ but go.mod declared the path without the required `/v2` suffix, so consumers cannot `go get` the v2 line.

- Rename module to `github.com/ysugimoto/falco/v2` in go.mod
- Rewrite all internal imports to the `/v2` path across the tree
- Update generator path literals so regenerated code stays consistent
- Update docs/README imports, `go install`, and pkg.go.dev links
- Retract mis-published v2.0.0-v2.3.0 (wrong module path)

Leave GitHub web URLs and the exception.go report URL unversioned.

Closes #579


acme:
We should probably try to get https://github.com/ysugimoto/falco/pull/593 and https://github.com/ysugimoto/falco/pull/614 in first, and merge them into this branch carefully.

jedisct1 (approved):

PR #616, opened 2026-06-18
fix: use http.NoBody for synthetic console/tester requests

mteodoro:
### Problem
Reading `req.body_bytes_read` (or `req.bytes_read`) in the `deliver`/`log` scopes panics when the request came from falco's synthetic constructors, because the getter calls `buf.ReadFrom(req.Body)` and `req.Body` is nil.

### Cause
Per the [`net/http.Request.Body` contract](https://pkg.go.dev/net/http#Request.Body), *"For server requests, the Request Body is always non-nil"* — so falco's real `ServeHTTP` path never yields nil. The nil arises only where falco builds requests with `http.NewRequest(..., nil)`: `NewRequest` leaves `Body == nil` when passed a nil body. `http.NoBody` is net/http's sentinel for an explicit empty body and yields a non-nil `Body`.

### Fix
Pass `http.NoBody` instead of `nil` at the four synthetic-request sites (`tester/tester.go`, `interpreter/console.go`) so they honor the same non-nil-Body contract as real server requests.

### Tests
- `examples/testing/request_byte_reads/` (in `TestTester`) drives the tester boundary via `call_subroutine("vcl_deliver"/"vcl_log")` reading `req.body_bytes_read`.
- `interpreter/console_test.go` `TestConsoleRequestBodyByteReads` drives the console boundary, same variable, both scopes.

Both panic without the fix.

### DIFF
--- a/cmd/falco/runner_test.go
+++ b/cmd/falco/runner_test.go
@@ -443,6 +443,12 @@ func TestTester(t *testing.T) {
 			filter: "*default.test.vcl",
 			passes: 2,
 		},
+		{
+			name:   "request byte-count reads with no body",
+			main:   "../../examples/testing/request_byte_reads/request_byte_reads.vcl",
+			filter: "*request_byte_reads.test.vcl",
+			passes: 2,
+		},
 	}
 
 	for _, tt := range tests {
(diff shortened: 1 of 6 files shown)


jedisct1:
AI-generated messages are always so warm and friendly.

jedisct1 (approved):

PR #618, opened 2026-06-24
feat(wasm): add include resolution and structured lint diagnostics

acme:
- Add in-memory `mapResolver` so `lint` resolves `include "..."` against a caller-supplied file map (matches CLI `-I`, `.vcl` suffix)
- Add `includes`/`mainFile` lint options; bare-string lint unchanged
- Expose parsed `AST` in the `lint` result alongside `errors`
- Add `File` origin attribution to each `LintError`
- Return parse failures as structured `errors` (line/position/file) instead of a flattened `error` string
- Always initialize `errors` as an array to avoid null on success
- Wire `pretest` to rebuild `falco.wasm`; fix `cp -f` in Makefile `wasm_exec` target; scope vitest to `tests/` and exclude node-only deps from optimization

Used by https://github.com/fastly/vscode-fastly-vcl/pull/102

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,4 +62,4 @@ wasm:
 
 wasm_exec:
 	@mkdir -p wasm
-	cp "$$(find "$$(go env GOROOT)" -name 'wasm_exec.js' 2>/dev/null | head -1)" wasm/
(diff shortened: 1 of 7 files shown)


acme:
I'm going with https://github.com/ysugimoto/falco/pull/622 instead

PR #620, opened 2026-06-24
feat: allow return(upgrade) in vcl_recv

acme:
- Add "upgrade" to valid RECV return states in the linter
- Fixes false error on WebSocket passthrough (`return(upgrade)`)
- Add linter test covering `return(upgrade)` in vcl_recv

Reference: https://www.fastly.com/documentation/reference/vcl/subroutines/recv/


ysugimoto (changes requested):
linter is OK but let's think about an interpreter side.
At least we need to specify `UPDATE` state in intepreter at https://github.com/ysugimoto/falco/blob/main/interpreter/state.go, in `stateMap`.

But of course local interpreter does not support WebSocket request (Fastly Fanout?) so we need to write a documentation about it.

PR #622, opened 2026-06-29
feat: add WASI component build target

acme:
- Add cmd/falco-component: reactor build exporting lint/format/ parse/tokenize via wit/falco.wit with hand-rolled canonical ABI
- Add resolver.MapResolver and token category helpers for the filesystem-less, string-in/string-out component API
- Add make wasm-component and make test-wasm targets, pinning the WASI preview1 adapter with a verified SHA-256
- Build and test the component in CI and attach it to releases via a shared setup-wasm-toolchain composite action
- Add host smoke tests (Python/JS/Ruby) and docs/wasm-component.md


acme:
This requires https://github.com/dip-proto/go-pcre/pull/1 to be merged and released

PR #623, opened 2026-06-29
feat: support week (w) RTIME unit

acme:
- Add `'w'` to the lexer's RTIME unit switch so `8w` lexes as a single RTIME token instead of `INT` + `IDENT`
- Handle `w` suffix in `ParseRTime` and in the interpreter (week = 24 × 7 hours), including `table.lookup_rtime`
- Fix misleading lexer comment: `m` is minute, not month
- Add lexer, parser, interpreter, and formatter tests covering RTIME units including week

Reference:
https://www.fastly.com/documentation/reference/vcl/types/rtime/

### DIFF
--- a/formatter/statement_format_test.go
+++ b/formatter/statement_format_test.go
@@ -215,6 +215,18 @@ func TestFormatSetStatement(t *testing.T) {
 				ExplicitStringConcat: true,
 			},
 		},
+		{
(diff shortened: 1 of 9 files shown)


acme:
I rolled this into https://github.com/ysugimoto/falco/pull/628

PR #624, opened 2026-06-29
fix: correct Fastly ACL matching semantics

acme:
- negated entries now exclude (longest-prefix match wins)
- "localhost" matches exactly 127.0.0.1 and ::1
- bare IPv6 entries default to /128
- non-IP string ~ acl returns false instead of erroring


ysugimoto (approved):
LGTM

PR #625, opened 2026-06-29
feat: add obj.is_hipaa variable

acme:
- Register `obj.is_hipaa` in supported scopes
- Return a fixed `false` value in the interpreter
- Add coverage for HIPAA and PCI placeholder behavior

Reference:
https://www.fastly.com/documentation/reference/vcl/variables/backend-response/beresp-hipaa/


ysugimoto (approved):
LGTM, thanks!

PR #626, opened 2026-06-29
fix: allow STRING implicit coercion in equality operators

acme:
- Permit `==`/`!=` between STRING and any STRING-coercible type (INTEGER/FLOAT/BOOL/TIME/IP) when the right operand is a constant, variable, or call, matching Fastly's left-driven coercion
- Reject numeric literals in this position, including those wrapped in grouping (`(10)`) or a prefix operator (`-10`)
- Keep coercion one-directional: non-STRING left operands must match exactly
- Add linter tests for coercion and rejected-literal cases
- Reference a Fiddle that verifies the behavior


ysugimoto (commented):
This implementation seems to be added in `main` branch already (maybe caused by my late review) but recursive evaluation of `GroupredExpression` in this PR is still worth to add.

PR #627, opened 2026-06-29
fix: resolve dotted HTTP header names in linter

acme:
- Treat the full path after `{req,resp,bereq,beresp,obj}.http.` as a single header name in `linter/context`, so headers like `req.http.one.two` no longer report `undefined variable`
- Consume all remaining dot-segments when a `%any%` slot is a leaf header container; keep per-segment resolution for nested slots (backend/director/ratecounter)
- Extract `resolveAnyKey` helper shared by `Get`/`Set`/`Unset` to centralize leaf-detection and key joining
- Add context unit tests and statement-linter integration tests for dotted header set/get/unset


akrainiouk (changes requested):
Looks like this implementation has problems linting the code like this:
```varnish
sub vcl_recv {
  #FASTLY RECV
  set req.http.foo = "foo";
  set req.http.foo.bar = "bar";
}
```
Leads to the following error:
```
                                                                                                                                                                                  
🔥 [ERROR] Type mismatch: req.http.foo.bar requires type NULL but STRING was assigned (operator/assignment)                                                                       
in /tmp/akrainio/main.vcl at line 4, position 24                                                                                                                                  
 3|  set req.http.foo = "foo";                                                                                                                                                    
 4|  set req.http.foo.bar = "bar";                                                                                                                                                
                          ^                                                                                                                                                       
 5|}                                                                                                                                                                              
See reference documentation: https://developer.fastly.com/reference/vcl/operators/#assignment-operators                                                                           
                                                                                                                                                                                  
🔥 2 errors, ❗ 0 warnings, 🔈 0 recommendations.                                                             ```
```

The fist statement injects `foo` item in http item list and then confuses the lookup when `foo.bar` is handled because it is treated as a parent object.

acme:
> Looks like this implementation has problems linting the code like this:

Thanks, good catch. https://github.com/ysugimoto/falco/pull/627/commits/652d6c3a78b6aa7269fc63b16f0f0852ef890feb adds that as a test and fixes that case.

akrainiouk (approved):
Looks good. Thanks!

PR #629, opened 2026-06-30
feat: widen beresp.backend.ip/name scope and resolve real IP

acme:
- Allow reading beresp.backend.ip/name in ERROR, DELIVER and LOG scopes, not just FETCH (matches Fastly behavior)
- Capture backend connection remote address via httptrace so beresp.backend.ip returns the real origin IP
- Extract nil-safe getBackendIP/getBackendName helpers and dedupe the fetch/deliver/error/log getters
- Return beresp.backend.ip as IP instead of String to match its declared type
- Add interpreter scope/no-backend and linter scope tests


ysugimoto (approved):
LGTM

PR #630, opened 2026-06-30
feat: support hex integer and exponent/hex float literals

acme:
- Lex hexadecimal integers (0x), decimal/hex float exponents ('e'/'p', lowercase only), and reject hex/exponent mantissas combined with an RTIME unit suffix
- Parse literals enforcing signed int64 range; allow 2^63 only under a unary minus (INT_MIN), rejecting masks and overflow
- Preserve the source literal through AST String(), the formatter, and codec encode/decode round-trips
- Read the evaluated integer value in the probe-health linter instead of re-parsing String()
- Document numeric literal forms in docs/parser.md and add lexer/parser/formatter/codec/interpreter tests


akrainiouk (approved):
Tested the behavior and it looks good.
As for the code, looks good to but there are many places this PR touches that I am not familiar with. So perhaps another pair of eyes would be good.

PR #631, opened 2026-07-01
feat: support fetch_timeout backend property

acme:
- Register `fetch_timeout` (RTIME) in the linter's `BackendPropertyTypes` so backend declarations using it pass validation
- Honor the `.fetch_timeout` property in the interpreter, bounding the fetch and overriding the `first_byte_timeout` default; the `bereq.fetch_timeout` variable still takes precedence
- Add positive and negative (wrong-type) linter tests for the property


akrainiouk (approved):
LGTM

PR #634, opened 2026-07-01
Fix parse_time_delta parser

jedisct1:
The parser was quite broken, for example 1d2h was parsed as 1 day + 12 hours.

We could fix it to return 1 day + 2 hours, but on Fastly servers, what follows the unit is silently ignored.

So, 1d2h parses fine, but returns the same duration as 1d.

So... do the same thing.


acme (approved):
Thanks! Verified with Fastly https://fiddle.fastly.dev/fiddle/3301394e

PR #636, opened 2026-07-05
fix: narrow {NAME} filter to match only .http.{NAME} variables

### LINKED ISSUE #287: [BUG] Documentation checker tool misses variables containing `{NAME}`
### Kind of proposals

- [x] Other

### Describe the problem

The documentation checking tool ([`cmd/documentation-checker/`](https://github.com/ysugimoto/falco/tree/main/cmd/documentation-checker) / `make check`) is setup to ignore all variables in the Fastly documentation pages that contain `{NAME}`. This is to ignore the header variables (`req.http.*`, etc). Which are defined in our [`predefined.yml`](https://github.com/ysugimoto/falco/blob/main/__generator__/predefined.yml) file in a different format than the upstream documentation (`req.http.%any%` vs `req.http.{NAME}`).

This prevented the tool from noticing [variables](#285) such as `backend.{NAME}.healthy`.

### Possible solution

This could be fixed by making header regex check more specific to match only the header variables or by removing the regex and instead converting the upstream format to match ours.

molloyzak13:
The documentation checker skips every variable containing `{NAME}` anywhere in its name. This was meant to filter only HTTP header variables (`req.http.{NAME}`, `bereq.http.{NAME}`, etc.) since Fastly docs use `{NAME}` as a placeholder but `predefined.yml` defines them as `req.http.%any%`. But `strings.Contains(name, "{NAME}")` also matches non-header variables like `backend.{NAME}.healthy`, which should be checked, not skipped.

Changed the filter to `strings.Contains(name, ".http.{NAME}")`. Every HTTP header variable in `predefined.yml` follows the `*.http.%any%` pattern, so requiring `.http.` before `{NAME}` still skips all five header groups (`req.http`, `bereq.http`, `beresp.http`, `obj.http`, `resp.http`) while non-header `{NAME}` variables now pass through.

Closes #287

### DIFF
--- a/cmd/documentation-checker/html.go
+++ b/cmd/documentation-checker/html.go
@@ -53,7 +53,7 @@ func fetchFastlyDocument(ctx context.Context, url string, m *sync.Map) error {
 			}
 			name := a.FirstChild.Data
 			if name == ignoreFunctionIf ||
-				strings.Contains(name, ignoreHTTPHeaderRelatedSignaure) ||
+				strings.Contains(name, ".http."+ignoreHTTPHeaderRelatedSignaure) ||
 				strings.Contains(name, ignoreRegexCapturedNumber) {
 
 				continue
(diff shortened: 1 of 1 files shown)


acme (approved):
Thanks!

acme (approved):
Thanks, I fixed the typo while I was there

Issue #297, opened 2024-04-05
[BUG] regex group variables reset on subroutine transitions


richardmarshall:
### Kind of proposals

- [x] Simulator
- [x] Testing

### Describe the problem

When changing scope to a new subroutine via `call` or a function expression the interpreter resets the `re.group.#` variables in the same way it does for local variables. This is not correctly following the behavior of Fastly.

Additionally accessing a re group variable before performing a regex match currently produces an undefined variable interpreter error instead of returning a `NotSet` value.

### VCL code that cause the problem / reproduceable

```vcl
sub foo {
  set req.http.foo1 = re.group.1;
  if (req.http.test ~ "([^.]+)") {}
  set req.http.foo2 = re.group.1;
}

sub bar {
  set req.http.bar1 = re.group.1;
  if (req.http.test ~ "(.*)") {}
  set req.http.bar2 = re.group.1;
  call foo;
  set req.http.bar3 = re.group.1;
}
```

```
  ●  [RECV] test

    Assertion Error: Assertion error: expect=aoeu.aoeu, actual=
    Actual Value: 

    5|   assert.equal(req.http.bar2, "aoeu.aoeu");
    6|   assert.equal(req.http.foo1, "aoeu.aoeu");
    7|   assert.equal(req.http.foo2, "aoeu");
```

### Expected behavior

https://fiddle.fastly.dev/fiddle/a3a1975b

ysugimoto:
Oh, I didin't know this spec...

I think we should follow the behavior in the interpreter but the linter should report an error as `WARNING` because this spec causes a potential bug.

richardmarshall:
@ysugimoto A linter warning is a good idea. I updated #294 to include a fix for the interpreter since I was already moving around the subroutine state changes. Can make another PR that adds warning logic to the linter.

PR #637, opened 2026-07-06
feat(linter): surface regex/uncaptured-variable rule for re.group.N reads

01luyicheng:
Resolves the **linter portion** of #297.

## Background

Per @ysugimoto's agreement in #297 (April 2024): the interpreter should follow Fastly semantics (regex groups persist across `call`), but the linter should WARN because relying on cross-subroutine scope leakage is a potential bug pattern. The interpreter side was already implemented by the maintainer in #412 (merged Feb 2025). This PR addresses the linter side, which was never completed.

## What I found

The `UncapturedRegexVariable` WARNING was **already functionally firing** — each subroutine is linted with a fresh regex-group state (`ResetRegexVariables()` on scope entry), so any `re.group.N` read before a match in that subroutine triggers the warning. This is exactly the cross-subroutine leak pattern from #297.

However, it had four gaps:

1. **Wrong rule name (bug):** the warning was matched to the `deprecated` rule via `err.Match(DEPRECATED)` instead of the dedicated `UNCAPTURED_REGEX_VARIABLE` rule (`regex/uncaptured-variable`) that already existed in `rules.go` but was unused. This made the warning mislabelled in output and un-ignorable by its proper name.
2. **No reference URL** for `UNCAPTURED_REGEX_VARIABLE` in the `references` map.
3. **No test coverage** for the #297 cross-subroutine `call` scenario (existing tests were single-subroutine only).
4. **No docs entry** for `regex/uncaptured-variable`.

## Changes

- **`linter/errors.go`** — Fix the rule mapping: `DEPRECATED` → `UNCAPTURED_REGEX_VARIABLE`. Improve the message to explicitly call out the cross-subroutine leak concern.
- **`linter/rules.go`** — Add reference URL pointing to issue #297 in the `references` map.
- **`linter/statement_linter_test.go`** — Add `TestRegexGroupedVariablesCrossSubroutine` covering the #297 `call` scenario:
  - `re.group.N` read before any match in a subroutine → WARNING (leak pattern)
  - Callee with its own match before the read → no warning
  - Match before read in the same subroutine → no warning
- **`docs/rules.md`** — Document the `regex/uncaptured-variable` rule with a VCL example and the `falco-ignore-next-line` silencing instruction.

## User-visible behavior change

The only breaking-ish aspect: anyone who previously silenced this WARNING via `# falco-ignore-next-line deprecated` (or `-ignore deprecated`) will now need `regex/uncaptured-variable`. This is an **intended** fix — the old `deprecated` mapping was a bug (it shared a rule with `DeprecatedVariable()`, making it impossible to selectively ignore the regex warning). The new dedicated rule name is the correct behavior.

No change to **when** the warning fires (verified conservative — no new false positives; matches in `if`/`else` branches correctly do not warn thanks to subroutine-global scope semantics).

Refs #297.

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -913,6 +913,23 @@ if (req.http.Host) {
 }
 ```
 
+## regex/uncaptured-variable
+
+A `re.group.N` variable is read without a preceding regex match (`~` or `!~`) in the current subroutine scope.
+
+`re.group.N` is subroutine-global in Fastly and its value can leak across `call` boundaries. Reading it before any match in the current subroutine means the value depends on a regex match performed by a caller (or is unset), which is a common source of subtle bugs. Falco reports this as a WARNING.
+
+```vcl
+sub foo {
+  // WARNING: re.group.1 read before any regex match in foo
+  set req.http.x = re.group.1;
+  if (req.http.test ~ "([^.]+)") {}
+  set req.http.y = re.group.1; // OK: preceded by a match
+}
+```
+
+To silence this warning for a specific line, use `# falco-ignore-next-line regex/uncaptured-variable`.
+
 ## disallow-empty-return
 
 A `return` statement in state-machine subroutine like `vcl_recv` must have the next state.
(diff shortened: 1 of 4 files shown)


acme:
Thanks, but please submit small, high-quality, fixes

PR #638, opened 2026-07-10
Fix: Seg faults when backend property is of a wrong type.

akrainiouk:
Currently backend property types are not enforced by the parser.
For example falco interpreter will accept backend port as an integer and instead of reporting an error will panic on backend request creation.

This PR simply adds necessary guards in createBackendRequest function so that it returns an error instead of panicking.
This PR does not try to fix the parser itself which might be worth doing as well.


acme (approved):
Thanks, looks good. I expanded it a little.

PR #639, opened 2026-07-23
fix(interpreter): in ==/!= operators allow type coercion from STRING to IP

akrainiouk:
Fastly allows to compare IP type with STRING as long as the right side is a variable or a literal representing valid IP value.

I noticed that type coercion rules are the same for assignment operator and for ==/!= and therefore utilized assign.Assign implementation instead of duplicating the logic.

I believe the same applies to remaining operand types so an entire Equal/NotEqual implementation can be improved by utilizing assign.Assign function in a similar way. I tried to limit the scope of this PR to IP type only to minimize the changes.
If this PR is accepted I will follow up with another PR covering STRING type with the goal to convert them all.


interpreter/assign/assign.go
```diff
@@ -239,11 +239,14 @@ func Assign(left, right value.Value) error {
...
-				return errors.WithStack(fmt.Errorf("invalid IP format, got %s", rv.Value))
-			} else {
+			if ip := net.ParseIP(rv.Value); ip != nil {
 				lv.Value = ip
 				lv.IsNotSet = false
+			} else if rv.IsLiteral() {
+				return errors.WithStack(fmt.Errorf("invalid IP format, got %s", rv.Value))
+			} else {
+				lv.Value = nil
+				lv.IsNotSet = true
```
akrainiouk, line 249: only reject invalid IP value for string literals.
For string variables invalid value is ignored and converted to uninitialized IP

interpreter/assign/assign_test.go
```diff
@@ -280,19 +280,19 @@ func TestProcessAssignment(t *testing.T) {
...
-		v := net.ParseIP("127.0.0.1")
-		vv := net.ParseIP("127.0.0.2")
+		v := value.IP{Value: net.ParseIP("127.0.0.1")}
+		vv := value.IP{Value: net.ParseIP("127.0.0.2")}
 		tests := []struct {
-			left    net.IP
+			left    value.IP
 			right   value.Value
-			expect  net.IP
+			expect  value.IP
```
akrainiouk, line 288: we need actual VCL values in order to check if the outcome is initialized or not.

acme (approved):
Thanks, looks good

PR #640, opened 2026-07-23
fix(interpreter): to string coercion in Equal/NotEqual brought in compliance with Fastly

akrainiouk:
following #639 the fix for STRING type coercion in ==/!= operator.
the code utilizes assign.Assign to convert right hand operand to STRING instead of duplicating the logic.


acme (approved):
Thanks, looks good. I made some small fixes while reviewing.

PR #642, opened 2026-08-02
fix: return a `not set` string when referencing an undefined field in an HTTP header

3846masa:
## Summary

This change fixes an issue where an undefined HTTP header field was treated as a defined empty string.

Fastly VCL handles string-to-boolean conversion differently from many programming languages. Only strings that are not set are falsy; a defined empty string is truthy.

Reference: [Fastly VCL best practices](https://www.fastly.com/documentation/guides/full-site-delivery/fastly-vcl/vcl-best-practices/)

## Problematic case

This means that VCL must distinguish between an HTTP header that is explicitly defined as an empty string and a field that is not defined within that header.

For example, when an unset field is read from an empty parent header, `req.http.Input:field-key` should be falsy:

```vcl
# https://fiddle.fastly.dev/fiddle/76847460

set req.http.Input = "";

# Expected: "field-key is not set" is logged
if (req.http.Input:field-key) {
  log "field-key is set";
} else {
  log "field-key is not set";
}
```

Before this fix, Falco treated `req.http.Input:field-key` as a defined empty string. Since defined empty strings are truthy, the condition evaluated to true.

On the other hand, the following case should return `true`, because the field itself is set even though its value is empty. Its boolean value should therefore be truthy:

```vcl
# https://fiddle.fastly.dev/fiddle/29a1b3b4

set req.http.Input = "field-key";

# Expected: "field-key is set to an empty string" is logged
if (req.http.Input:field-key) {
  log "field-key is set to an empty string";
} else {
  log "field-key is not set";
}
```

## Root cause

In the current Falco implementation, an empty HTTP header value is always handled first.

Falco checks whether the header itself is unset, then returns the empty value.

This also happens when Falco reads a field value from the HTTP header.

As a result, when the header value is explicitly defined as an empty string, an undefined field was also returned as a defined empty string.

In Fastly VCL, an undefined field must be returned as an unset string.

## Changes

The code now separates whole-header lookups from field-value lookups.

Whole-header lookups are used when no field key is specified.

Field-value lookups always call `GetField()`.

This produces the intended behavior for empty header values.

### DIFF
--- a/cmd/falco/runner_test.go
+++ b/cmd/falco/runner_test.go
@@ -411,7 +411,7 @@ func TestTester(t *testing.T) {
 			name:   "header subfield dealing test",
 			main:   "../../examples/testing/subfield-header-dealing/default.vcl",
 			filter: "*default.test.vcl",
-			passes: 10,
+			passes: 24,
 		},
 		{
 			name:   "synthetic vs obj.response",
--- a/examples/testing/subfield-header-dealing/default.test.vcl
+++ b/examples/testing/subfield-header-dealing/default.test.vcl
@@ -78,3 +78,107 @@ sub test_recv {
     set req.http.VARS:VALUE = req.http.NULL;
     assert.equal(req.http.VARS, "VALUE");
 }
+
+// Reading a subfield distinguishes "not set" from a defined empty string.
+// Fastly only treats not-set strings as falsy, so an absent subfield of a
+// header that is itself defined must still read as not set.
+// All test cases referred to Fastly fiddle behaviors
+// see: https://fiddle.fastly.dev/fiddle/28404963
+
+// @scope: recv
+// @suite: GET UNDEFINED VARS VALUE FROM EMPTY VARS
+sub test_recv {
+    declare local var.state STRING;
+    set req.http.VARS = "";
+    if (req.http.VARS:VALUE) {
+        set var.state = "truthy";
+    } else {
+        set var.state = "falsy";
+    }
+    assert.equal(var.state, "falsy");
+    assert.is_notset(req.http.VARS:VALUE);
+}
+
+// @scope: recv
+// @suite: GET EMPTY VARS ITSELF
+sub test_recv {
+    declare local var.state STRING;
+    set req.http.VARS = "";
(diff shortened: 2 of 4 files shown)


acme:
:wave: Nice find. I poked at this on the edge a bit more than the two fiddles above, and Fastly backs you up on all of it:

https://fiddle.fastly.dev/fiddle/28404963

Checked absent subfield of an empty header (falsy), the empty header itself (truthy), a bare `VALUE` with no `=` (truthy), `VALUE=` (truthy), parent header never set (falsy), different key present (falsy).

The one I'd flag that isn't in your description: `set req.http.COPY = req.http.VARS:VALUE;` used to leave COPY *set*, since the bogus empty string propagated on assignment. Fastly leaves it unset. Your patch fixes that too. I built header.go at its pre-patch state to check, and that case plus the plain read are the only two that flip.

On tests though: the Go unit tests only hit the helpers, so nothing runs this through actual VCL. `examples/testing/subfield-header-dealing/` already covers this corner and runs in CI, and `examples/testing/notset/` does the whole "fiddle link + VCL test" thing. Threw the cases in there in that style in case you want them:

https://github.com/acme/falco/tree/acme/subfield-notset-tests

Just appends to `default.test.vcl` and bumps passes 10 → 24, nothing else. Two of them fail without your fix. Cherry-pick `35592ad`, adapt it, ignore it, whatever's easiest. I couldn't push to your branch directly, maintainer edits are off on this one.

Oh, and it wants a rebase, it's behind main.

3846masa:
@acme 

Thank you for your review.

> I couldn't push to your branch directly, maintainer edits are off on this one.

Because I forked the repository using an organization account, I wasn’t able to grant you edit access. Instead, I’ve temporarily added you as a contributor to the forked repository.

> Cherry-pick 35592ad, adapt it, ignore it, whatever's easiest.
> Oh, and it wants a rebase, it's behind main.

I think your suggestion to add VCL tests was excellent. As you suggested, I cherry-picked 35592ad and incorporated it. I’ve also rebased the branch onto main.

acme (approved):
Thanks, looks good

PR #643, opened 2026-08-04
fix(interpreter): query string space encoding made consistent with Fastly

akrainiouk:
querystring changed to encode space characters as %20 instead of "+" which makes it consistent with Fastly


acme:
Thanks for this, you've got the diagnosis right, and I checked it against real edge nodes rather than trusting the docs: Fastly really does emit `%20` ([05e9f66c](https://fiddle.fastly.dev/fiddle/05e9f66c), [83f6cc79](https://fiddle.fastly.dev/fiddle/83f6cc79), [b22c048e](https://fiddle.fastly.dev/fiddle/b22c048e)), escaping the key is right too, and `queryEscape` handles the fiddly case nicely, since `url.QueryEscape` turns a literal `+` into `%2B` before the `ReplaceAll` ever gets to it.

I had a little think about where the escaping lives, though, and I'm not sure `String()` is the right home for it. Fastly leaves parameters it doesn't touch exactly as it found them, byte for byte, so escaping on the way out rewrites things it shouldn't: `sort("/p?b=1%202&a=x+y")` comes back as `/p?a=x+y&b=1 2` on Fastly but `/p?a=x%20y&b=1%202` here, and `set({"/p?e=1%202&f=a+b&g=x y"}, "a", "z")` ought to leave `f` and `g` well alone.

Those `+` cases happened to round-trip correctly before this change, so I think we'd be swapping one set of mismatches for another rather than clearing them out.

There's a third form hiding in here too, which caught me out: a plain `"…%20…"` literal pct-decodes at parse time, so the function actually receives a raw space, and Fastly passes that straight through as well, and since `ParseQuery` collapses `+`, `%20` and a raw space into the same string, nothing we do in `String()` can tell them apart afterwards.

So I think we need a rethink, based on much Fiddle testing so that we match Fastly behaviour.

PR #644, opened 2026-08-14
fix: align querystring builtin behaviour

acme:
- add support for `querystring.sort(..., only_unique_keys)`
- preserve raw encoding, parameter order, duplicates, and valueless parameters
- align sorting, filtering, cleaning, add, and set behavior
- enforce the 32-parameter processing limit
- add regression coverage for query string edge cases
- semantics verified against Fastly Fiddle

### DIFF
--- a/interpreter/function/builtin/querystring_sort.go
+++ b/interpreter/function/builtin/querystring_sort.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/ysugimoto/falco/v2/interpreter/context"
 	"github.com/ysugimoto/falco/v2/interpreter/function/errors"
-	"github.com/ysugimoto/falco/v2/interpreter/function/shared"
 	"github.com/ysugimoto/falco/v2/interpreter/value"
 )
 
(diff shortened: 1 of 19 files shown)


interpreter/function/builtin/querystring_sort.go
```diff
@@ -35,14 +57,128 @@ func Querystring_sort(ctx *context.Context, args ...value.Value) (value.Value, e
...
+		sorted = append(sorted, qsToken{})
+		copy(sorted[p+1:], sorted[p:])
+		sorted[p] = tok
+	}
+	return sorted
+}
+
+// qsEmit joins tokens after removing only leading empty entries.
+func qsEmit(prefix string, tokens []qsToken) string {
+	p := 0
```
ysugimoto, line 173: let's put comment for keeping the last token even when empty - this is intentional (covered in testing).

ysugimoto (approved):
LGTM with nit

interpreter/function/builtin/querystring_sort.go
acme: Thanks, added in https://github.com/ysugimoto/falco/pull/644/commits/f8a6f26825c03d317198bc6d3d1257a4b4653e57

PR #645, opened 2026-08-16
fix: parse IPv4 and IPv6 addresses for client.ip correctly

3846masa:
# Summary

Fix an issue where `client.ip` returned `nil` when the simulator received a request from an IPv6 client.

# Background

In the simulator, `client.ip` is determined from the client address in `RemoteAddr`.
For IPv6 connections, `RemoteAddr` uses a bracketed format such as `[2001:db8::1]:12345`.

The previous implementation split the address at the last `:`, so it could not correctly parse IPv6 addresses.

# Changes

- Use `net.SplitHostPort` to extract the host portion from `RemoteAddr`
- Support IPv4 and IPv6 values for `client.ip`
- Apply the same parsing logic to `client.identity`
- Add regression tests covering IPv4, IPv6, and addresses with and without ports

# Tests

The following test command passes:

```text
go test ./interpreter/variable
```

# Scope of Impact

Only the way `client.ip` and `client.identity` are resolved in the simulator is changed.
The default values in testing and console modes are not affected.

### DIFF
--- a/interpreter/variable/all.go
+++ b/interpreter/variable/all.go
@@ -535,11 +535,11 @@ func (v *AllScopeVariables) Get(s context.Scope, name string) (value.Value, erro
 		}
 		if v.ctx.ClientIdentity == nil {
 			// default as client.ip
-			idx := strings.LastIndex(req.RemoteAddr, ":")
-			if idx == -1 {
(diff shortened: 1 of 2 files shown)


ysugimoto (approved):
LGTM (sorry for the late)

PR #646, opened 2026-08-21
feat: add utf8.translate function

jedisct1:
Implement the utf8.translate built-in and move the utf8.* functions to the correct documentation category by the way ("unicode").

On fastly, the Unicode functions set fastly.error, so do that as well:
- utf8.stdpad can report EUTF8 for broken input, and EDOM for the smallest integer width
- and utf8.substr returns a not-set string when the request falls outside the string, so do that too.


ysugimoto (changes requested):
Overall looks good, but I cannot find actual `builtin.Utf8_translate(ctx, args...)` function implementation in this PR. Did you add like `utf8_translate.go` with test files?

PR #659, opened 2026-09-11
Report a recursive include instead of resolving it forever

### LINKED ISSUE #658: [BUG] An include loop is resolved until the process runs out of memory
### Kind of proposals

- [x] Linter
- [x] Simulator

### Describe the problem

An include statement that names a module which is already being resolved is resolved again, forever. `resolveIncludeStatements` in `linter/linter.go` and `resolveIncludeStatement` in `interpreter/include.go` both recurse without recording which modules are already on the include path, so a module that includes itself, directly or through the modules it includes, is loaded and parsed until the process runs out of memory.

Honest about how I found it: a script of mine generated an entry point inside the directory it was globbing, so the generated file included itself. falco grew to 9.9GB RSS over 17 minutes, took all 7GB of swap on a 15GB machine and the shell session died with it. My mistake, but a linter should say so rather than take the machine down.

`main.vcl`:

```vcl
include "deps";

sub vcl_recv {
  #FASTLY RECV
  set req.http.Foo = "bar";
}
```

`deps.vcl`:

```vcl
include "deps";
```

```
$ ulimit -v 2000000  # so this does not cost you your machine
$ falco lint main.vcl
runtime: out of memory: cannot allocate 4194304-byte block (78381056 in use)
fatal error: out of memory
```

Two files including each other do the same:

```vcl
// a.vcl
include "b";

// b.vcl
include "a";
```

So does a symlink, where the names differ but the file is the same:

```
$ ln -s real.vcl link.vcl   # real.vcl contains: include "real";
$ falco lint main.vcl       # main.vcl contains: include "link";
```

The simulator takes the same path through `interpreter/include.go`. The formatter does not resolve includes, so `falco fmt` is not affected.

### Expected behavior

One error naming the loop, and the rest of the file still linted.

- OS: Amazon Linux 2023
- Shell: bash

### Additional context

PR: #659

With that patch the first case above reports

```
[ERROR] Module deps is included recursively: /tmp/repro/deps.vcl -> /tmp/repro/deps.vcl (include/module-recursion)
```

in 0.03s at 20MB RSS, and the symlink case names both the link and the file it points at.

zfLQ2qx2:
Fixes #658.

Both include resolvers recurse with no record of where they have been, so a module that includes itself is loaded until the memory runs out. This keeps the include path on the way down and reports the statement that closes the loop.

`resolver.IncludeStack` is the record. A module is identified by the name its resolver returned and, when that name is a file, by the file itself through `os.SameFile`, so a symlink and its target are one module rather than two, and the reported path names both. Resolvers that do not read from disk still get the check on the name, which is enough to break the loop. There is no depth limit, the identity is the whole mechanism.

`linter` reports `include/module-recursion` at ERROR and drops that one include, so the rest of the file is still linted. `interpreter` raises a runtime error naming the loop. The same module included twice in different places is not a loop and still resolves twice, which the tests pin down.

Measured, self include, mutual include and a symlink pair: out of memory before, one error and exit 1 after, 0.03s at 20MB RSS. Output over the 264 VCL files I have here and every VCL under `examples/` is byte identical to main. Suite is green.

@ysugimoto


interpreter/include.go
```diff
@@ -22,11 +22,21 @@ func (i *Interpreter) resolveIncludeStatement(statements []ast.Statement, isRoot
...
+			// includes, would be resolved until the process runs out of memory.
+			if !i.includes.Push(module) {
+				return nil, exception.Runtime(
+					&stmt.GetMeta().Token,
+					"VCL module '%s' is included recursively: %s -> %s",
+					include.Module.Value, i.includes.Path(), module,
+				)
+			}
 			recursive, err := i.resolveIncludeStatement(included, isRoot)
+			i.includes.Pop()
```
ysugimoto, line 39: linter calls `IncludeStack.Pop()` with `defer` keyword so I suggest that this function should do the same:

```suggestion
			i.includes.Pop()
			recursive, err := i.resolveIncludeStatement(included, isRoot)
```

This gives a bit more guarantee for future logic added after `i.resolveIncludeStatement(included, isRoot)` — `i.includes.Pop()` will certainly be called and pop the include stack.

ysugimoto (approved):
Overall looks good to me, just left a comment with a tiny suggestion.

PR #648, opened 2026-09-04
Add fmt error-on-unformatted option

ccouzens:
Closes https://github.com/ysugimoto/falco/issues/647

This makes falco easier to use in CI scripts, where you're interested if the changed code meets the conventions.

PR #662, opened 2026-09-12
cmd/falco: run every file the fmt command was given

### LINKED ISSUE #655: fmt gives up on the remaining files after the first one that fails
`falco fmt` takes a list of files, and the loop over the resolvers in
`cmd/falco/main.go` breaks on the first error. A file that does not parse costs every
file listed after it, so argument order decides how much work gets done.

Three files. `a.vcl` and `b.vcl` are indented with 4 spaces and format fine.
`broken.vcl` is missing a semicolon.

```
$ falco fmt -w broken.vcl a.vcl b.vcl
Missing semicolon
in broken.vcl at line 2, position 22
$ echo $?
1
```

`a.vcl` and `b.vcl` are untouched. Move the bad file to the end and they are not:

```
$ falco fmt -w a.vcl b.vcl broken.vcl
Formatted a.vcl.
Formatted b.vcl.
Missing semicolon
in broken.vcl at line 2, position 22
$ echo $?
1
```

Same files, same flags, same exit status, different amount of work done. Nothing in
the first run says it stopped early.

`fmt` is the only subcommand affected. It is built from `NewGlobResolver` over every
argument, so it gets one resolver per file, where `lint`, `simulate`, `stats` and
`test` go through `NewFileResolvers` and always get exactly one.

This matters more with #647 and #648. A check mode that stops at the first failure
under-reports: over a tree of 264 `.vcl` files, `falco fmt -x` names 1 file as needing
formatting when 207 of them do.

PR: #662. It sits on top of the #648 branch, because two of its tests use the `-x`
flag and so need #648 first, and it says so. If you would rather have this land on
its own I can rebase it onto main with `-w` only tests.

zfLQ2qx2:
Fixes #655.

**Depends on #648. Do not merge this first.** It is written as if #648 is already
in: the branch is #648's head plus one commit, so until #648 merges this diff
shows their commit too. Two of the tests here drive the `-x` flag that #648 adds.
Once #648 lands I will rebase and this becomes one commit against main.

You raised the same thing reviewing #648: "With --check, if one file needs
formatting, the command stops and remaining files are not checked." That is what
this fixes, for `-w` as well as `-x`.

### The cause

`main` builds one resolver per file for `fmt`, then breaks out of the loop the
first time an action returns `ErrExit`:

```go
if exitErr == ErrExit {
    shouldExit = true
    break
}
```

Every other subcommand goes through `NewFileResolvers` and gets exactly one
resolver, so the `break` is invisible there. `fmt` goes through
`NewGlobResolver` per argument, so it stops at the first file that fails and
leaves the rest untouched: unformatted with `-w`, unreported with `-x`. Nothing
in the output says it stopped early, and the exit status is the same either way.

Over a tree of 264 .vcl files, `falco fmt -x` names 1 file as needing formatting
when 207 of them do.

### The fix

Drop the `break` and keep the flag, so the exit status is unchanged and every
file is visited. The loop moves into `runResolvers` so it can be tested without
`os.Exit`.

### Tests

`TestFormatVisitsEveryFile`, three cases: `-w` formats the files after one that
does not parse, `-x` reports every file that needs formatting, and `-x` stays
quiet and succeeds when they are all formatted. Each builds the files in a temp
dir and runs the real command through `runResolvers`.

```
--- PASS: TestFormatVisitsEveryFile/-w_formats_the_files_after_one_that_does_not_parse
--- PASS: TestFormatVisitsEveryFile/-x_reports_every_file_that_needs_formatting
--- PASS: TestFormatVisitsEveryFile/-x_says_nothing_and_succeeds_when_every_file_is_formatted
```

The same commit is also offered on ccouzens/falco#1, in case it is easier for it
to ride along with #648 rather than land after it. Either way is fine by me.

@ysugimoto

### DIFF
--- a/cmd/falco/help.go
+++ b/cmd/falco/help.go
@@ -215,8 +215,9 @@ Usage:
     falco fmt [flags] ...files
 
 Flags:
-    -h, --help         : Show this help
-    -w, --write        : Overwrite format result
+    -h, --help                 : Show this help
+    -x, --error-on-unformatted : Return a non zero exit code if formatting is required
+    -w, --write                : Overwrite format result
 
 files argument accepts glob file patterns
 
--- a/cmd/falco/runner.go
+++ b/cmd/falco/runner.go
@@ -561,6 +561,18 @@ func (r *Runner) Format(rslv resolver.Resolver) error {
 
 	formatted := formatter.New(r.config.Format).Format(vcl)
 	var w io.Writer
+	if r.config.Format.ErrorOnUnformatted {
+		formatted_buf := new(strings.Builder)
+		_, err := io.Copy(formatted_buf, formatted)
+		if err != nil {
+			return errors.WithStack(err)
+		}
+		formatted_string := formatted_buf.String()
+		if main.Data != formatted_string {
+			return errors.WithStack(fmt.Errorf("%s requires formatting", main.Name))
+		}
+		formatted = strings.NewReader(formatted_string)
+	}
 	if r.config.Format.Overwrite {
 		writeln(cyan, "Formatted %s.", main.Name)
 		fp, err := os.OpenFile(main.Name, os.O_TRUNC|os.O_WRONLY, 0o644)
--- a/config/config.go
+++ b/config/config.go
@@ -99,7 +99,8 @@ type ConsoleConfig struct {
 // Format configuration
 type FormatConfig struct {
 	// CLI options
-	Overwrite bool `cli:"w,write" default:"false"`
+	Overwrite          bool `cli:"w,write" default:"false"`
+	ErrorOnUnformatted bool `cli:"x,error-on-unformatted" default:"false"`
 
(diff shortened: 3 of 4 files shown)


ccouzens:
Nice to meet you zfLQ2qx2,

I've been making changes to https://github.com/ysugimoto/falco/pull/648 based on ysugimoto's feedback. So far my changes have just been variable names, option names and the formatting that comes as a result.

You may want to take in the same changes so that this PR merges cleanly on top of it.

---

https://github.com/ysugimoto/falco/pull/648#pullrequestreview-5182807394

> ## the behavior for multiple file input
> 
> The fmt subcommand accepts multiple files including glob patterns. With --check, if one file needs formatting, the command stops and remaining files are not checked. In CI, users would expect the command to report all files that need formatting before exiting with an error.

I will point ysugimoto to this PR for this.

